PR #56, opened 2023-05-16
Update Publication Model and Allow Primary/Secondary Publications

bencap:
This change updates our publication model to remove implications that all publication must be PubMed publications. It also adds API support for the enumeration of whether a publication associated with an experiment or scoreset should be identified as a primary publication or just a standard associated publication.

To do this, two new association objects (one each for scoresets and experiments), were added to our models so that SQLAlchemy can handle an association table with an additional column to the two primary keys. 

Along with this change, we may consider refactoring existing entries in the database table to denote whether existing publication identifier associations are primary or secondary.


afrubin:
@bencap This all looks good to me, thanks!

Am I correct in understanding that all existing publications will be flagged as not primary when this is added? It's fine if that is the case, and it looks like primary publications can be designated by editing existing score sets using the API.

Is it enforced that a record can only have one primary publication? I'm not sure that we need to enforce this strictly in the back end but either way is fine.

It looks like the API output will have a `"publicationIdentifiers"` list and a separate `"publicationIdentifierAssociations"` that would have to be joined by the user or client consuming the API output. That's not a problem and I think your creation of the association table is the right choice, but we should think about whether there's a straightforward way to combine these in the API output so it's nicer to deal with.

bencap:
@afrubin Yes, that is all correct. 

For existing publications, I can add in database migration logic that would set a publication as primary if there was only one publication. For experiments/scoresets with multiple publications, we would need to handle any changes manually.

It is not enforced at the database level that a experiment have only one primary publication, and the data model would allow it. However, our Pydantic view models do validate that an incoming primary publication identifiers list has a maximum of one element. From line 59 in `view_models/experiment.py`:
```
primary_publication_identifiers: conlist(PublicationIdentifierCreate, min_items=0, max_items=1)
```
So we should never get anything back from the UI that causes an experiment or scoreset to have multiple primary publications. This would be easily changed depending on how many primary publications you want to limit a scoreset/experiment to.

Right, and thats an artifact of this implementation that I don't love. The view components of the UI rely exclusively on `publicationIdentifierAssociations`, since those contain the publication list in addition to whether or not each publication has been marked `primary`. It's a bit messier for our edit components though, since they have to use the `publicationIdentifierAssociations` list to populate our existing data in the UI but still return separate `publicationIdentifiers` and `primaryPublicationIdentifiers` lists to the API. I've been trying to think about a better way to handle that operation, but what I have tried so far hasn't quite worked.

PR #63, opened 2023-05-24
Mapped variant

EstelleDa:


EstelleDa:
There are a folder mappings and a script in the Zip folder. Mappings is a sample data folder. Script for inserting mapped variants in database is a Jupyter Notebook file. 
I use Anaconda to open a Jupyter Notebook. 
Installing some packages got some problems. Details are in the script.
I use alembic to produce tables in local database first. This script is only used in inserting data to local database. 
 
[data and script.zip](https://github.com/VariantEffect/mavedb-api/files/11607408/data.and.script.zip)

src/mavedb/routers/mapped_variant.py
```diff
@@ -0,0 +1,44 @@
...
+
+@router.get("/{urn}", status_code=200, response_model=mapped_variant.MappedVariant, responses={404: {}, 500: {}})
+async def show_mapped_variant(
+    *, urn: str, db: Session = Depends(deps.get_db)
+) -> Any:
+    """
+    Fetch a mapped variant by URN.
+    """
+
+    return await fetch_mapped_variant_by_variant_urn(db, urn)
```
bencap, line 44: I think it makes more sense to just have the logic within `fetch_mapped_variant_by_variant_urn` here instead of wrapping it inside a separate function, similar to how we fetch existing experiments and scoresets via URN here: https://github.com/VariantEffect/mavedb-api/blob/ef48654cbc8fc7dd4649e8b0716ac2f60789f5f0/src/mavedb/routers/experiments.py#L75-L84.

That way we can keep these router files clean of any non-router processing functions. If you wanted to retain a separate function, I would create a new file `src/mavedb/variants.py` and put the function there. I think this is more applicable to search though. Again, from the experiments router: https://github.com/VariantEffect/mavedb-api/blob/ef48654cbc8fc7dd4649e8b0716ac2f60789f5f0/src/mavedb/routers/experiments.py#L65-L72

bencap (approved):
Looks good @EstelleDa. Couple non-critical comments about code organization but functionally all good.

bencap:
I suppose since we merged the changes to the publication identifiers you also might need to rebase onto the dev branch and change the alembic downgrade target.

src/mavedb/routers/mapped_variant.py
```diff
@@ -0,0 +1,44 @@
...
+    :param urn: The variant URN.
+    :return: The mapped variant, or None if the URL was not found or refers to a private score set not owned by the specified
+        user.
+    """
+    try:
+        variant = db.query(Variant).filter(Variant.urn == urn).one_or_none()
+        if not variant:
+            raise HTTPException(status_code=404, detail=f"Mapped variant with URN {urn} not found")
+        item = db.query(MappedVariant).filter(MappedVariant.variant_id == variant.id).one_or_none()
+    except MultipleResultsFound:
```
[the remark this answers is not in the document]
EstelleDa: Thanks Ben! I didn't think combine these two filters together when I wrote this function. I only thought if no variant then no need to keep searching. I'll combine these two together cause it'll be faster.
For `MultipleResultsFound`, my understanding is a urn should only match one variant and a variant should have only one mapped variant. This is why throwing an error if there are more than one mapped variant with the same urn.

src/mavedb/routers/mapped_variant.py
EstelleDa: I'm not sure what new tasks Samriddhi will do for Mapped Variants so that I leave it in a separating file. Fetching existing scoreset is in routers/scoreset.py. I prefer to move it to routers/scoreset.py rather than routers/experiment.py if there's only one fetching mapped variants function because mapped variants link to scoreset variants.

bencap: Oh yeah no worries, it makes sense in `scoresets.py`. I was just giving an example of how when we do searching in `experiments.py` we have a separate helper function for the search (`_search_experiments`). I think that would be applicable here too. Instead of having two separate functions, we could just combine them into one:
```
@router.get("/{urn}", status_code=200, response_model=mapped_variant.MappedVariant, responses={404: {}, 500: {}})
async def show_mapped_variant(
    *, urn: str, db: Session = Depends(deps.get_db)
) -> Any:
    """
    Fetch a mapped variant by URN.
    """
    try:
        variant = db.query(Variant).filter(Variant.urn == urn).one_or_none()
        if not variant:
            raise HTTPException(status_code=404, detail=f"Mapped variant with URN {urn} not found")
        item = db.query(MappedVariant).filter(MappedVariant.variant_id == variant.id).one_or_none()
    except MultipleResultsFound:
        raise HTTPException(status_code=500, detail=f"Multiple variants with URN {urn} were found.")
    if not item:
        raise HTTPException(status_code=404, detail=f"Mapped variant with URN {urn} not found")
    return item
```

We likely won't need to reuse the fetch logic outside of the router, so I think it's ok to just have one function in the router file that handles all the fetch logic.

EstelleDa: I think it depends on whether this function will be used more than one time. If only `show_mapped_variant` needs it, it's better to combine them such as `fetch_experiment` in routers/experiment.py. Otherwise, it's good to separate them such as `fetch_scoreset_by_urn` in routers/scoreset.py. Currently MappedVariant is a draft model. We'll change this model's attributes and add more functions when Samriddhi join us.

bencap: Agreed, makes sense to me.

src/mavedb/routers/mapped_variant.py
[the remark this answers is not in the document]
EstelleDa: I never think this problem before. We can talk more about this with Jeremy next meeting. `fetch_scoreset_by_urn` in 
`routers/scoreset.py` and `show_user` in `routers/users.py` throw `MultipleResultsError` too if there're multiple results. 
p.s. We can modify error information mavedb-ui side, but some part doesn't work very well. I'm not sure how's going now cause at that time we were modifying many validators. Hence I stopped doing that.

EstelleDa:
> I suppose since we merged the changes to the publication identifiers you also might need to rebase onto the dev branch and change the alembic downgrade target.
I changed the alembic `down_revision = 'da9ba478647d'`. Is it enough or do I need to use alembic to reproduce this file?

bencap:
> > I changed the alembic `down_revision = 'da9ba478647d'`. Is it enough or do I need to use alembic to reproduce this file?
That should be enough since none of your changes are affected by the most recent database changes.

EstelleDa:
> > I changed the alembic `down_revision = 'da9ba478647d'`. Is it enough or do I need to use alembic to reproduce this file?
> 
> That should be enough since none of your changes are affected by the most recent database changes.

Great. Thanks Ben!

jstone-uw:
I've changed the base of this pull request from `dev` (now obsolete) to `release-2023.1.3`. There's one conflict, which I'll resolve.

src/mavedb/routers/scoresets.py
```diff
@@ -174,6 +177,25 @@ async def get_scoreset_counts_csv(
     return StreamingResponse(iter([stream.getvalue()]), media_type="text/csv")
 
 
+@router.get(
+    "/scoresets/{urn}/mappedVariants", status_code=200, response_model=list[mapped_variant.MappedVariant])
```
afrubin, line 178: Should be `mapped-variants` to be consistent with other API endpoints right?

EstelleDa: You're right. I'll change it.

ahwagner:
Looking forward to this! @afrubin is aware, but just making a note that we will be referring to the `mapped-variants` endpoint for the the mapping paper. I'll follow along here but let me know if this changes.

EstelleDa:
> Looking forward to this! @afrubin is aware, but just making a note that we will be referring to the `mapped-variants` endpoint for the the mapping paper. I'll follow along here but let me know if this changes.

Hi @ahwagner, I changed it already. The branch is above your comment,  [ff19262](https://github.com/VariantEffect/mavedb-api/pull/63/commits/ff19262446d305c7a50eadf5d22e9ec7bbd40984).

PR #65, opened 2023-06-06
Support Preprints as Valid Publications

bencap:
Support preprints within the MaveDB API


src/mavedb/server_main.py
jstone-uw, line 1: I wonder if we might want to wrap `InvalidPMID` in our own exception class at some point.

src/mavedb/view_models/experiment.py
jstone-uw, line 1: "Publication identifiers lists must be a collection" is a little cryptic. I would suggest either
- "Publication identifier lists must be collections"
- Or (probably the better option) get the name of the actual field being validated. I think this can be done by adding a third validator parameter, `info: FieldValidationInfo`, and then accessing `info.field_name`. I haven't tried it, but see the source code at https://github.com/pydantic/pydantic-core/blob/f38bea601a5b81d73ee5a132242b7b97d24772ff/python/pydantic_core/core_schema.py.

tests/helpers/util.py
jstone-uw, line 1: Do we still want to log the response?

bencap: Nope!

src/mavedb/view_models/experiment.py
bencap: Thanks for the tip, it seems that the relevant info is actually provided when you define a fourth key word argument, and then we can simply ignore the values keyword.

PR #67, opened 2023-06-07
Enable Experiment/Scoreset Searches via Associated Publication Identifiers

bencap:


src/mavedb/lib/experiments.py
jstone-uw, line 1: Is there a reason not to search experiments by exact publication identifier, like we do for score sets?

src/mavedb/routers/publication_identifiers.py
jstone-uw, line 1: Should any of the publication API endpoints be protected from abuse by requiring the user to be authenticated? It's not a strong protection since anyone can register, but maybe it's worthwhile for any endpoint that isn't used on the search screen.

Also, I'd change the function name fetch_experiment to fetch_publication, even though it doesn't have practical consequences.

jstone-uw (changes requested):

src/mavedb/routers/publication_identifiers.py
bencap: Yeah that makes sense and I'll go ahead and protect them. I think the idea here is that at some point these endpoints can be used to populate the initial set of filter options at some point instead of fetching all the score-sets. That isn't implemented at the moment, but at that point we could remove the protections.

Thanks for catching the function name, I keep copy pasting routes and making that same mistake...

src/mavedb/lib/experiments.py
bencap: Thanks, I thought I had added it but I did not...

src/mavedb/routers/publication_identifiers.py
bencap: In light of this I added another fetch route for just identifiers and made those routes raise a `MultipleResultsFound` exception. That exception was previously raised by the search endpoints, which now will just return a list of all matching publications.

PR #70, opened 2023-06-15
Assign meta-analysis score sets to special experiments not created by…

jstone-uw:
An experiment URL must be provided when the score set is not a meta-analysis, and it must not be provided when the score set is a meta-analysis.

**Changes**

- An experiment URL must be provided when the score set is not a meta-analysis, and it must not be provided when the score set is a meta-analysis.
- When creating a meta-analysis, the API looks for any existing meta-analysis with the same set of underlying source score sets.
  - If one exists, it assigns the new meta-analysis to the same experiment.
  - Otherwise, it creates a new experiment using the score set's title, short description, abstract, and methods (but not its extra JSON metadata).

### DIFF
--- a/src/mavedb/lib/score_sets.py
+++ b/src/mavedb/lib/score_sets.py
@@ -4,7 +4,7 @@ import numpy as np
 import pandas as pd
 from pandas.testing import assert_index_equal
 from sqlalchemy import func, or_
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, aliased
 
 from mavedb.lib.array_comparison import assert_array_equal
 from mavedb.lib.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


jstone-uw:
Some conditions can arise that seem unusual at first glance. For instance, consider this scenario:

1. User A creates meta-analysis tmp:1 for score sets X and Y, in new experiment tmp:a.
2. User B creates another meta-analysis, tmp:2, for the same score sets X and Y. It gets assigned to the same unpublished experiment tmp:a.
3. User B publishes score set tmp:2.

Experiment tmp:a is now published as urn:1, with one published score set (tmp:2, now urn:1.1). Experiment 1 was created by user A, whose score set is still unpublished.

I don't think this results in any real data inconsistency; user A can still publish tmp:1 as score set urn:1.2. The result is that we never find two experiments containing meta-analyses for the same collection of underlying score sets, as desired. But it may be worthwhile to hide a meta-analysis experiment's creator, since by design it may include score sets created by multiple users.

jstone-uw:
The validation described above occurs in a Pydantic root validator, i.e. one that is not associated with a specific field. As a result, validation errors for missing experiments will not currently cause the UI to highlight the experiment input. A UI update is needed to display errors not associated with particular fields, to support this and other use cases.

PR #77, opened 2023-07-25
Replace Metapub with Eutils

bencap:
Metapub is not as well supported as it was in the past and uses a better maintained package,
eutils, under the hood. This change replaces previous functionality that used Metapub to
search for PubMed articles with eutils.


bencap:
We will need to add an `NCBI_API_KEY` environment variable to our environment when we deploy this change.

PR #81, opened 2023-09-08
Add Support for Genomic Coordinate Score Sets

bencap:


jstone-uw:
Here are a couple of minor things I noticed, which you might want to address now or later.

1. If something is missing from a gene target when the score set is submitted, the "Required" error message appears inside the collapsed target details, so the user may not be able to tell what went wrong.

Initial state (collapsed):
Screenshot 2023-10-16 at 6 57 16 PM

Expanded:
Screenshot 2023-10-16 at 6 57 11 PM

2. When searching by accession or transcript identifier, `{}` is shown in the drop-down options if no assembly or HGNC name has been chosen yet.

Screenshot 2023-10-16 at 6 57 37 PM

jstone-uw:
Would you let me know what target(s) to use with the SGE scores and counts? Sorry, you might have mentioned it somewhere.

bencap:
> Would you let me know what target(s) to use with the SGE scores and counts? Sorry, you might have mentioned it somewhere.

CARD11 is the gene and NM_032415.5 is the accession, for anyone else who happens to take a look at this.

Issue #92, opened 2023-10-17
Smaller API responses for score and count previews


jstone-uw:
To display a preview of scores and counts in a score set, the UI currently fetches `/api/v1/score-sets/{urn}/scores` and `/api/v1/score-sets/{urn}/counts`. These endpoints return complete score and count data, but we only need a small number of rows for the preview. Pagination parameters should be added so that the client can request something like `/api/v1/score-sets/{urn}/scores?start=0&limit=100`.

EstelleDa:
@jstone-uw I created a branch for this one and merging it to main branch needs at least one reviewer to approve. Let me know if anything you wanna change to it.

PR #93, opened 2023-11-08
Add pagination parameters in previous scores and counts

EstelleDa:
Solve #92


src/mavedb/routers/score_sets.py
```diff
@@ -153,7 +165,17 @@ def get_score_set_scores_csv(
 
     columns = ["accession", "hgvs_nt", "hgvs_splice", "hgvs_pro"] + score_set.dataset_columns["score_columns"]
     type_column = "score_data"
-    rows_data = get_csv_rows_data(score_set.variants, columns=columns, dtype=type_column)
+    variants = score_set.variants
+
```
ashsny, line 169: something like 
variants = variants[start if start else 0:]
if limit:
  variants = variants[:limit]

is a little cleaner imo

src/mavedb/routers/score_sets.py
```diff
@@ -175,12 +197,24 @@ def get_score_set_scores_csv(
...
     *,
     urn: str,
+    start: int = Query(default=None, description="Start index for pagination"),
+    limit: int = Query(default=None, description="Number of variants to return"),
     db: Session = Depends(deps.get_db),
     user: User = Depends(get_current_user),
 ) -> Any:
     """
     Return counts from a score set, identified by URN, in CSV format.
+    If no start and limit, all of variants of this score set will be returned.
```
ashsny, line 207: comment says variants, should probably be counts

src/mavedb/routers/score_sets.py
```diff
@@ -175,12 +197,24 @@ def get_score_set_scores_csv(
 async def get_score_set_counts_csv(
     *,
     urn: str,
+    start: int = Query(default=None, description="Start index for pagination"),
+    limit: int = Query(default=None, description="Number of variants to return"),
```
ashsny, line 201: number of counts?

src/mavedb/routers/score_sets.py
EstelleDa: Thanks for your feedback. I tried it. It can't handle correctly if both `start` and `limit` are present. The reason is `variants` is cut three times. The condition still needs clarify if users only type either `start` or `limit`.

> something like if start: variants = variants[start:] if limit: variants = variants[:limit]
> 
> is a little cleaner imo

EstelleDa: @ashsny In my email I see your comments “comment says variants, should probably be counts”, but I don’t see it in GitHub. I don't know why. 
I leave comments of both `limit` of scores and counts are number of variants. The reason is both of them belong to variants. Some variants have both and some have only scores but no count data. The `limit` number returns the number of variants, but not the number of counts or scores. I don't think we need part of counts or scores. That will be huge amount of data again. It's easy to understand their relationship in Table `variants` Column `data`.

ashsny: I'm not sure what you mean by "`variants` is cut three times." Is there a specific case you think won't be handled correctly by the code I posted? Maybe I am mis-understanding the API here. 

see examples:
https://colab.research.google.com/drive/17tEMiKutF8Wrh2AnA9rdWBQ7mR7lXbGz?usp=sharing

EstelleDa: Sorry, my fault. I thought these codes are still there. 
```
if start and limit:
        end = start + limit
        variants = variants[start:end]
```
Yes, yours is better. I'll change these parts and push again.

ashsny:
You probably know a lot more about this than me, but is there any way to change the pull request so it is limited to the changes in the last two commits? There are a bunch of older commits included that I'm not sure should be merged into main (I would expect only the two most recent commits to be merged into something like release-2023.5.0 or release-2023.4.1 or something).

EstelleDa:
> You probably know a lot more about this than me, but is there any way to change the pull request so it is limited to the changes in the last two commits? There are a bunch of older commits included that I'm not sure should be merged into main (I would expect only the two most recent commits to be merged into something like release-2023.5.0 or release-2023.4.1 or something).

Good point. I didn't notice that. I'm not sure how's going with those older commits. Maybe we can talk about it in next dev meeting and think a better way to merge codes together.

PR #95, opened 2023-11-08
Fix ongoing 500s and broken homepage search links by updating target organism search .to match new schema.

ashsny:
This PR both changes the currently broken target_organism_names search, which is creating 500s/alerts, and also adds target_organism_names search to the top-level search, which fixes the Homo sapiens etc. links on the homepage (https://github.com/VariantEffect/mavedb-ui/issues/64).


ashsny:
Fixes #96 and https://github.com/VariantEffect/mavedb-ui/issues/64.

PR #102, opened 2023-11-16
Enforce unique URNs at schema level for URN-having entities.

ashsny:
These include experiment_sets, experiments, scoresets, and variants.

I went through the db and manually removed existing duplicates.


jstone-uw:
Looks good.

My only suggestion is to append a descriptive string to the Alembic migration filename, for future readability. Alembic doesn't care about the filenames, so this can be done manually. There's also a way to accomplish it with Alembic's -m option. (https://alembic.sqlalchemy.org/en/latest/autogenerate.html)

ashsny:
> Looks good.
> 
> My only suggestion is to append a descriptive string to the Alembic migration filename, for future readability. Alembic doesn't care about the filenames, so this can be done manually. There's also a way to accomplish it with Alembic's -m option. (https://alembic.sqlalchemy.org/en/latest/autogenerate.html)

Done.

PR #118, opened 2023-12-15
Add a checked-in JSON file of our OpenAPI.

ashsny:
Pending a merging of `mavedb-api` and `mavedb-ui`, in order to provide type safety in the frontend, we will temporarily check in this file and instructions for updating it. The `mavedb-ui` repo will then transform this to typescript types with the `openapi-typescript` Node package. This process will eventually become more automated and less silly, but for now this will suffice.


jstone-uw (approved):

ashsny:
I think we want to abandon this approach in favor of grabbing json from the running dev server. It's much easier and faster, although it does require a running server. For a hopefully temporary process, I believe it is more appropriate.

PR #121, opened 2023-12-20
Optimize SQL queries for score set searches.

### LINKED ISSUE #120: Optimize score set search queries.
Calls to search_score_sets result in a large number of queries, because many (or all?) relationships are being lazy-loaded, and they are being loaded one by one as SQLAlchemy model objects are transformed into Pydantic objects.

jstone-uw:
Fixes #120

Calls to search_score_sets have been resulting in a large number of queries, often lazy-loading individual relationships separately for each row in the search results. Here we use SQLAlchemy's eager-loading options to reduce the number of queries.

With this change, calls to search_score_sets involve
- One query that loads matching score sets, together with all of the relationships that will be returned in the ShortScoreSet view model;
- And a series of subqueries for all the score sets belonging to the experiments from the first query, together with all of the necessary relationships.

The subqueries use SELECT IN to keep the number of queries small -- currently, three queries cover all published experiments.

We might revisit whether all of the relationships in ShortScoreSet are really necessary for search results.

### DIFF
--- a/src/mavedb/lib/score_sets.py
+++ b/src/mavedb/lib/score_sets.py
@@ -2,10 +2,9 @@ from typing import Optional
 
 import numpy as np
 import pandas as pd
-import json
 from pandas.testing import assert_index_equal
-from sqlalchemy import func, or_, and_
-from sqlalchemy.orm import Session, aliased
+from sqlalchemy import func, or_
+from sqlalchemy.orm import aliased, contains_eager, joinedload, selectinload, Session
 
 from mavedb.lib.array_comparison import assert_array_equal
 from mavedb.lib.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


jstone-uw:
Here's a log of the queries run by this new version for a search with no filters:

[query_log.txt](https://github.com/VariantEffect/mavedb-api/files/13730096/query_log.txt)

src/mavedb/lib/score_sets.py
```diff
@@ -127,7 +131,50 @@ def search_score_sets(db: Session, owner: Optional[User], search: ScoreSetsSearc
         )
 
     score_sets: list[ScoreSet] = (
-        query.join(ScoreSet.experiment).join(ScoreSet.target_genes).order_by(Experiment.title).all()
+        query.join(ScoreSet.experiment)
```
ashsny, line 134: is it intentional that the top-level `.join(ScoreSet.target_genes)` goes away here?

ashsny:
For the record, do you have a rough estimate of how much this speeds up the empty search query?

PR #122, opened 2023-12-21
Nick/add GitHub ci

nickzoic:
Adds github workflow to run pytest tests automatically on push.

Some small modifications have been made to get tests to pass & these need to be checked to make sure I'm not "fixing" them in the wrong direction.

Work in progress:

- [X] Running pytest
- [ ] Coverage Report
- [x] Python 3.11 version doesn't cache pandas wheel so is super slow.

### DIFF
--- a/src/mavedb/lib/identifiers.py
+++ b/src/mavedb/lib/identifiers.py
@@ -149,7 +149,7 @@ class ExternalPublication:
         else:
             doi_str = "" if not self.preprint_doi else self.preprint_doi
             title = "(None)" if not self.title else self.title.strip(".")
-            journal = "(None)" if not self.publication_journal else self.publication_journal.strip(".")
+            journal = "(None)" if not (hasattr(self, "publication_journal") and self.publication_journal) else self.publication_journal.strip(".")
             year = "(Unknown year)" if not self.preprint_date else self.preprint_date.year
 
             # We don't receive these fields from rxiv platforms
--- a/.github/workflows/run-tests-on-push.yml
+++ b/.github/workflows/run-tests-on-push.yml
@@ -12,5 +12,5 @@ jobs:
       with:
         python-version: "3.11"
     - run: pip install --upgrade pip
-    - run: pip install .[dev]
+    - run: pip install .[dev,server]
     - run: PYTHONPATH=. pytest tests/validation/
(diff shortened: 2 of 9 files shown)


src/mavedb/lib/identifiers.py
```diff
@@ -149,7 +149,7 @@ def reference_html(self) -> str:
         else:
             doi_str = "" if not self.preprint_doi else self.preprint_doi
             title = "(None)" if not self.title else self.title.strip(".")
-            journal = "(None)" if not self.publication_journal else self.publication_journal.strip(".")
+            journal = "(None)" if not (hasattr(self, "publication_journal") and self.publication_journal) else self.publication_journal.strip(".")
```
ashsny, line 152: I must admit this change mystifies me a bit. It reads identically to the previous code for me. What am I missing?

ashsny (approved):
Looks good to me!

src/mavedb/lib/identifiers.py
nickzoic: the "publication_journal" attribute doesn't always exist, which means checking "not self.publication_journal" throws an AttributeError.  `hasattr(self, "publication_journal")` checks that the attribute exists ... it might still be None or otherwise false but at least we know it exists before we try to check its value.

This is ... an unusual way to do things.
I think it would be better to initialize it (and other attributes) to None if it isn't appropriate, but I don't know if that would affect anything else in the code so I just did it this way as a workaround.
@bencap might have some input here?

bencap: I didn't initialize those fields to make it clear in the `ExternalPublication` object that those fields were not defined for a particular publication on some external source of data, rather than just empty. The `ExternalPublication` model only ever acts as a temporary container when converting between some returned information from the PubMed / ___rxiv APIs. I had believed it would be more useful to raise an error when trying to access those fields and that `reference_html` issue got past me.

If we were to initialize those attributes to `None` instead, it shouldn't effect any other parts of the codebase. Places that would try to access a non-existent attribute are getting some sort of blanket default at the moment (https://github.com/VariantEffect/mavedb-api/blob/149c101457cb179062caf728382441852f6edbf1/src/mavedb/lib/identifiers.py#L304C19-L304C19).

nickzoic:
Update:

* I still can't get coverage to work: it always reports 0% coverage for src/ files.  It works perfectly on my local machine, of course.
* I switched on caching for pip packages which improves test times a great deal
* mypy / pylint are going to take a lot of work to fix the code.

I'll continue looking into this but I reckon it's worth merging what we've got now and I'll make a new PR for the outstanding stuff.

nickzoic:
Adding mypy is now it's own issue: #126

PR #128, opened 2024-01-11
Expand Genomic Scoreset Transcript Pool

bencap:
- Uses the CDOT package to fetch transcript sequences instead of built-in biocommons UTA methods in order to expand our pool of scoresets.
- Adds both samtools and htslib to our Docker image. These tools are used to index the GRCh37 and GRCh38 genome builds that CDOT relies on for sequence fetching of the expanded transcript pool.
- Adds a new dep `hgvs_data_provider`. This dep exposes a `RESTDataProvider` that provides access to our expanded seqpool. An instance of `ChainedSeqFetcher` is used by this object, which will default to fetching sequences from SeqRepo if they exist, but fall back on the new GRCh37 and GRCh38 data if no sequence matching the transcript exists in SeqRepo.
- This new dep is used wherever we previously accessed biocommons HGVS to ensure we always use the same pool of transcripts and sequences.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,40 @@
+FROM python:3.9 AS downloader
+
+WORKDIR /data
+
+# Install tools necessary used to install samtools and htslib so we can configure fasta files for genomic assembly.
+RUN apt-get update && apt-get install -y \
+	build-essential \
+	curl \
+	git \
+	libbz2-dev \
+	libcurl4-openssl-dev \
+	libgsl0-dev \
+	liblzma-dev \
+	libncurses5-dev \
+	libperl-dev \
+	libssl-dev \
+	zlib1g-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install samtools and htslib.
+ARG htsversion=1.19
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,5 +1,38 @@
 FROM python:3.9
 
+WORKDIR /data
+
+# Install tools necessary used to install samtools and htslib so we can configure fasta files for genomic assembly.
+RUN apt-get update && apt-get install -y \
```
nickzoic, line 6: If we don't need these tools when running, it might be worth doing this as a [multi-stage build](https://docs.docker.com/build/building/multi-stage/), just so all that build-essential type stuff isn't left behind in the final container.  It's a pretty small change to what you've already got, just add in: 
```FROM python:3.9 AS downloader```
up the top and then
```FROM python:3.9
COPY --from=downloader /data /data```
before `WORKDIR /code`.

bencap: Oh cool, didn't even know about that- thanks for the tip! I'll try it out tomorrow AM.

PR #135, opened 2024-01-19
Tests & Fixups for /api/v1/hgvs endpoints

nickzoic:
This PR adds tests for the /api/v1/hgvs/ endpoints, adds a couple of exception handlers in endpoints:

* `.../{assembly}/accessions`
* `.../fetch/{accession}`

It also removes `response_model = list` from some endpoints which returned a single entry, which gets the tests to pass.
If we actually want to return a single item list, we should fix that though:

* `.../transcripts/gene/{gene}`
* `.../transcripts/{transcript}`
* `.../transcripts/protein/{transcript}`

### DIFF
--- a/src/mavedb/routers/hgvs.py
+++ b/src/mavedb/routers/hgvs.py
@@ -7,6 +7,7 @@ import bioutils.assemblies
 from cdot.hgvs.dataproviders import RESTDataProvider
 from fastapi import APIRouter, Depends, HTTPException
 from hgvs import parser, validator
+from hgvs.exceptions import HGVSDataNotAvailableError
 
 from mavedb.deps import hgvs_data_provider
 
@@ -22,7 +23,10 @@ def hgvs_fetch(accession: str, hdp: RESTDataProvider = Depends(hgvs_data_provide
     """
     List stored sequences
     """
-    return hdp.seqfetcher.fetch_seq(accession)
+    try:
+        return hdp.seqfetcher.fetch_seq(accession)
+    except HGVSDataNotAvailableError as e:
+        raise HTTPException(404, str(e))
 
 
 @router.post("/validate", status_code=200, response_model=Any)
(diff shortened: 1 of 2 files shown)


src/mavedb/routers/hgvs.py
```diff
@@ -104,15 +111,15 @@ def list_transcripts_for_gene(gene: str, hdp: RESTDataProvider = Depends(hgvs_da
...
-@router.get("/transcripts/{transcript}", status_code=200, response_model=list)
+@router.get("/transcripts/{transcript}", status_code=200)
 def transcript_info(transcript: str, hdp: RESTDataProvider = Depends(hgvs_data_provider)):
     """
     List transcript information for a particular transcript
     """
     return hdp.get_tx_identity_info(transcript)
 
 
-@router.get("/transcripts/protein/{transcript}", status_code=200, response_model=str)
```
ashsny, line 115: reminder to check this out

src/mavedb/routers/hgvs.py
```diff
@@ -104,15 +111,15 @@ def list_transcripts_for_gene(gene: str, hdp: RESTDataProvider = Depends(hgvs_da
...
+@router.get("/transcripts/protein/{transcript}", status_code=200)
```
nickzoic, line 122: (recheck this)

src/mavedb/routers/hgvs.py
nickzoic: yeah if this is declared with `response_model=str` then two tests fail with:
```
pydantic.error_wrappers.ValidationError: <exception str() failed>
```
when serializing the response.  This is because `hdp.get_pro_ac_for_tx_ac` can return either a string (if there's a matching protein accession) or None (if there isn't).  So instead we're returning a JSON string (eg: with quotes around it) or a JSON null.

The mavedb-ui vue function `swapNucleotideProteinAccessions` I think expects this, or at least it passes `accept: application/json` so I assume that it expects JSON back.

nickzoic: Adding to the confusion, I think maybe axios, if it asks for JSON but doesn't get JSON back, just hands over the raw string received as `response.data`,

PR #141, opened 2024-01-25
Rebased taxonomy

EstelleDa:
Very messy changes. Rebase with main branch first and then merge with release-2023.5.0 branch. 

Original relationship: GenomeIdentifier-ReferenceGenome-ReferenceMap-TargetGene
New relationship: GenomeIdentifier-Taxonomy-TargetSequence-TargetGene

Mavedb-ui branch is newTaxonomy.


nickzoic (commented):
My only question is: are the `op.execute("""UPDATE` commands in the alembic migration sufficient to cleanly update all existing records from the old schema to the new one?  It seems odd that they're picking out individual tg.scoreset_ids for update rather than a transform approach.

EstelleDa:
> My only question is: are the `op.execute("""UPDATE` commands in the alembic migration sufficient to cleanly update all existing records from the old schema to the new one? It seems odd that they're picking out individual tg.scoreset_ids for update rather than a transform approach.
The update commands are produced by another codes I wrote in Jupyter Notebook. They're based on a file that Alan gave to me before. It's a good point that whether we have some more new score sets in current database and this file covers all score sets. I confirm private score sets are not in the list. 
[published_score_set_tax_ids.csv](https://github.com/VariantEffect/mavedb-api/files/14092195/published_score_set_tax_ids.csv)

EstelleDa:
Alan replies to me that csv file doesn't include the whole score sets we have now. Maybe we can merge this branch after I getting the whole data and updating the update commands in alembic file.

src/mavedb/lib/score_sets.py
```diff
@@ -93,10 +93,9 @@ def search_score_sets(db: Session, owner: Optional[User], search: ScoreSetsSearc
 
     if search.target_organism_names:
         query = query.filter(
-            ScoreSet.target_genes.any(
-                TargetGene.target_sequence.has(
-                    TargetSequence.reference.has(ReferenceGenome.organism_name.in_(search.target_organism_names))
-                )
+            ScoreSet.target_gene.target_sequence.has(
```
ashsny, line 96: why isn't this target_genes.any( as before?

src/mavedb/routers/target_genes.py
```diff
@@ -77,11 +77,13 @@ def search_target_genes(search: TextSearch, db: Session = Depends(deps.get_db))
 
     if search.text and len(search.text.strip()) > 0:
         lower_search_text = search.text.strip().lower()
+        print(lower_search_text)
```
ashsny, line 80: please remove all print() debugging from this file

src/mavedb/routers/taxonomies.py
```diff
@@ -0,0 +1,131 @@
...
+    db: Session = Depends(deps.get_db),
+) -> Any:
+    """
+    List distinct species names, in alphabetical order.
+    """
+
+    items = db.query(Taxonomy).all()
+    concatenates = []
+    for i in items:
+        if i.common_name != 'NULL' and not None:
```
ashsny, line 28: what is the `and not None` part of this clause doing? should it be `i.common_name != 'NULL' and i.common_name is not None`?

src/mavedb/routers/taxonomies.py
```diff
@@ -0,0 +1,131 @@
...
+from mavedb import deps
+from mavedb.models.taxonomy import Taxonomy
+from mavedb.view_models import taxonomy
+from mavedb.view_models.search import TextSearch
+
+router = APIRouter(
+    prefix='/api/v1/taxonomies', tags=['taxonomies'], responses={404: {'description': 'Not found'}}
+)
+
+@router.get("/concatenateTaxonomyList", status_code=200, response_model=List[str], responses={404: {}})
```
ashsny, line 16: why do we need this API call? why is it called `concatenateTaxonomyList` specifically? how is this intended to be used?

src/mavedb/routers/taxonomies.py
```diff
@@ -0,0 +1,131 @@
...
+            lower_search_text = search.text.strip().lower()
+            query = query.filter(
+                or_(
+                    func.lower(Taxonomy.organism_name).contains(lower_search_text),
+                    func.lower(Taxonomy.common_name).contains(lower_search_text)
+                )
+            )
+        else:
+            query = query.filter(Taxonomy.tax_id == int(search.text))
+    # It leads dropdown button doesn't work in Mavedb-ui autocomplete component.
```
ashsny, line 124: remove commented-out code

src/mavedb/lib/score_sets.py
EstelleDa: typo

src/mavedb/routers/taxonomies.py
EstelleDa: I shall delete this function. It was originally planned for used in dropdown menu, but later I noticed that I could fulfil it in UI side directly.

ashsny:
mavedb-api change looks good but has conflicts that must be resolved in lib/score_sets.py

src/mavedb/lib/score_sets.py
```diff
@@ -62,8 +62,8 @@ def search_score_sets(db: Session, owner: Optional[User], search: ScoreSetsSearc
                 ScoreSet.keyword_objs.any(func.lower(Keyword.text).icontains(lower_search_text)),
                 ScoreSet.target_genes.any(
                     TargetGene.target_sequence.has(
-                        TargetSequence.reference.has(
-                            func.lower(ReferenceGenome.organism_name).icontains(lower_search_text)
+                        TargetSequence.taxonomy.has(
+                            func.lower(Taxonomy.organism_name).contains(lower_search_text)
```
ashsny, line 66: I'm not sure you want to change icontains to contains here

src/mavedb/lib/taxonomies.py
```diff
@@ -0,0 +1,61 @@
...
+    If it exists, saving it in local database and return it to users.
+    """
+    if search and len(search.strip()) > 0:
+        if search.isnumeric() is False:
+            search_text = search.strip().lower()
+        else:
+            search_text = str(search)
+
+        url = f"https://api.ncbi.nlm.nih.gov/datasets/v2alpha/taxonomy/taxon/{search_text}"
+        async with httpx.AsyncClient() as client:
```
ashsny, line 35: when we fetch pubmed articles from NCBI, we don't use httpx but rather Eutils with our NCBI API key. I think you should probably use that here so we don't run into issues being rate limited from using the free api.

link: https://github.com/VariantEffect/mavedb-api/blob/89945b5310d2eacf1ce391398e3b8d298e5810e4/src/mavedb/lib/identifiers.py#L182

alembic/versions/6fae83d65ee4_taxonomy.py
```diff
@@ -0,0 +1,4646 @@
+"""taxonomy
```
ashsny, line 1: is this script final? have you tested the most recent version?

src/mavedb/lib/score_sets.py
EstelleDa: My branch may mix with previous codes. It's interesting that the above filters all used `contains`. I changed this to `icontains`.

EstelleDa: Wait. No, in main branch all of them are `contains`. I don't know which part modifies them in this PR. 
https://github.com/VariantEffect/mavedb-api/blob/main/src/mavedb/lib/score_sets.py

alembic/versions/6fae83d65ee4_taxonomy.py
EstelleDa: Yes, it's the final one. I rolled it back to the previous one. The foreign key of taxonomy in target sequence is taxonomy.id, not taxonomy.tax_id. I used this one to rebuild my local database. Special Taxonomy detail is https://github.com/VariantEffect/mavedb-api/issues/155. It's used to handle the taxonomy is NULL. Most of those score sets are unpublished or invalid.

src/mavedb/lib/taxonomies.py
EstelleDa: I might need time to do some searches about it. Taxonomy seems in EInfo.

src/mavedb/lib/score_sets.py
ashsny: you should be looking at branch release-2024.0.0, not main, in which it is icontains. this was a change made in https://github.com/VariantEffect/mavedb-api/commit/17f9671df662f8f25efa5a0f810e2354b19229c2

requirements.txt
```diff
@@ -10,6 +10,7 @@ fastapi~=0.71.0
 fqfa~=1.2.3
 IDUtils~=1.2.0
 hgvs~=1.5.4
+httpx~=0.24.1
```
ashsny, line 13: remove this when you remove the usage below

src/mavedb/lib/score_sets.py
EstelleDa: Yes, you're correct.

src/mavedb/lib/taxonomies.py
EstelleDa: - Eutils doesn't support taxonomy database. 
- EInfo only gets database information. 
- ESearch only supports term search. It neither supports tax_id search nor return all of necessary values we need. 
- Biopython Entrez.efetch only returns taxonomy's taxId and ScientificName. It's mainly used in getting GenBank data. 
- ELink can be used to find related items in the NCBI Entrez databases. Only getting related ID lists. I don't understand what these related IDs are for taxonomy. PubMed can get information that if a paper has been cited. 

NCBI doesn't appear to have a way to create an NCBI API key that is tied to "MaveDB" and not a person with a user account. - Alan

I think getting NCBI taxonomy in original way is fine. Not so many new taxonomies are found every year. It's different from lot of new publication papers published every year. The dropdown menu list covers the common taxonomies should work for most cases so usually MaveDB doesn't need to send a request to NCBI.

src/mavedb/view_models/target_sequence.py
```diff
@@ -13,7 +13,8 @@ class TargetSequenceBase(BaseModel):
     sequence_type: str
     sequence: str
     label: Optional[str]
-    reference: ReferenceGenome
+    # Without this one targetGene.targetSequence?.taxonomy.organismName in front end doesn't work
+    taxonomy: Taxonomy
```
bencap, line 17: Since Taxonomy is a required property of these target sequences and the alembic migration script is a data migration that fills in taxonomies for each target via `id`, we will just want to make sure that data migration is up to date for all score sets in the db before running the migration.

Alternatively, we could do something like
```
op.execute("""UPDATE target_sequences SET taxonomy_id = 0 WHERE taxonomy_id is NULL""")
```
as the last step of the migration, just to ensure any residually empty target sequences are filled with the empty Taxonomy.

EstelleDa: Target sequence will throw a validating error and leads to Search page doesn't work if taxonomy is null or empty. I think it's a good idea to prevent any new data that doesn't have taxonomy in the future.

EstelleDa: Target sequence will throw a validating error and lead to Search page doesn't work if taxonomy is null or empty. I think it's a good idea to prevent any new data that doesn't have taxonomy in the future.

jstone-uw:
I'm interested in what others think, but I would suggest that if possible, we avoid using Alembic migrations that refer to particular records/IDs, because
(a) In a new installation of MaveDB, these queries would be meaningless at best and might fail;
(b) And they're based on the state of the database at a particular point in time, so they need to be revised if new score sets are created before the update is released.

So I'm wondering if it's possible to rewrite 6fae83d65ee4_taxonomy.py to use more general queries. I'm guessing there are clear rules for deriving these INSERT and UPDATE statements from existing records. I'd be happy to help rework the migration too, if @EstelleDa can share the rules that were used to generate the current queries.

EstelleDa:
> I'm interested in what others think, but I would suggest that if possible, we avoid using Alembic migrations that refer to particular records/IDs, because (a) In a new installation of MaveDB, these queries would be meaningless at best and might fail; (b) And they're based on the state of the database at a particular point in time, so they need to be revised if new score sets are created before the update is released.
> 
> So I'm wondering if it's possible to rewrite 6fae83d65ee4_taxonomy.py to use more general queries. I'm guessing there are clear rules for deriving these INSERT and UPDATE statements from existing records. I'd be happy to help rework the migration too, if @EstelleDa can share the rules that were used to generate the current queries.

I did three times comparisons. The first time was based on current MaveDB website data. The second time is comparing the missing target sequences with local database. And comparing any different value in these two steps. Final time filtering any target sequences don't have any reference genome in local database. I think had better merge and update database with these new values ASAP. Otherwise, more and more values might be meaningless when users change something or more new values are created in the online MaveDB website.
Welcome any other suggestions for this part.

PR #148, opened 2024-01-30
Nick/138 mock requests

nickzoic:
This moves tests to running in isolation with no contact to the outside world.
This is enforced by the `pytest-socket` library which intercepts all socket calls.
Tests which would normally contact CDOT or NCBI are then mocked out.

It also makes test results substantially more verbose which is handy for understanding failures when looking at the CI output.

### DIFF
--- a/tests/routers/test_hgvs.py
+++ b/tests/routers/test_hgvs.py
@@ -1,3 +1,9 @@
+import requests_mock
+
+import cdot.hgvs.dataproviders
+from hgvs.exceptions import HGVSDataNotAvailableError
+from unittest.mock import patch
+
 VALID_MAJOR_ASSEMBLY = "GRCh38"
 VALID_MINOR_ASSEMBLY = "GRCh38.p3"
 INVALID_ASSEMBLY = "undefined"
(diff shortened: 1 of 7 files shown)


tests/routers/test_hgvs.py
```diff
@@ -79,67 +103,158 @@ def test_hgvs_genes(client, setup_router_db):
...
-    assert response.status_code == 200
-    assert response.json() is None
+    with requests_mock.mock() as m:
+
+        m.get(
+            'https://cdot.cc/gene/fnord',
+            status_code=404,
+        )
+        response = client.get(f"/api/v1/hgvs/genes/{INVALID_GENE}")
+        # XXX this probably SHOULD return a 404, but currently returns a 200 with None
```
ashsny, line 130: want to file an issue, TODO(#issue) all of these, and merge as is? or is this WIP?

nickzoic: I can raise -api and -ui issues for the 5 `XXX` items there ... it's not that they're *wrong* per se it's jut that a GET for a thing which doesn't exist would normally by convention return a 404.

Plenty of other APIs break this convention though so I'm okay with not worrying about it.

Issue #126, opened 2024-01-11
Code checks with mypy


nickzoic:
mypy is a very handy analysis tool which we should be using ... it'll take a bit of wrangling to get our current code to pass but it does tend to find interesting bugs (eg: #125) especially once we get typing support from our dependencies (eg: https://github.com/VariantEffect/mavehgvs/pull/36 )

nickzoic:
now the various other branches are merged this work will continue ( https://github.com/VariantEffect/mavedb-api/tree/nick/126-code-checks-with-mypy )

nickzoic:
OK so there's one tricky thing here, which is that there's lots of changes to models, specifically to relationship declarations, which are hard to parse out from the diffs.

So I wrote [a script](https://github.com/VariantEffect/mavedb-api/blob/nick/126-code-checks-with-mypy/tests/dump_rels.py) which dumps the details of relationships before and after a change and using that I've been able to make sure no relationships got lost and no options change.  Plus, like, the tests pass.

bencap:
Closing this as completed since we now have working code checks with mypy. We could probably expand what we have typed vs. untyped, but that can be a separate issue.

PR #156, opened 2024-02-20
Nick/126 code checks with mypy

nickzoic:
OK this adds code checks with mypy as per #126 

Summary of changes:
* Added in `mypy`, various stub libraries and some mypy specific configuration to pyproject.toml
* Added in a github action to run the mypy tests automatically
* Added type stubs (all the `*.pyi` files under `mypy_stubs`) for libraries which don't or won't add their own
* Removed `lib/array_comparison.py` which was very difficult to get types working in and replaced it with a much simpler pandas function `arrays_equal`.
* Removed the `lib/worker.py` and related stuff.  If we ever need it again, we can dredge it out of revision control.
* Changed a lot of SQLAlchemy declarations to the [latest style](https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html) to make mypy happy.
* Wrote a `tests/dump_rels.py` which dumps out a list of entity relationships and their option flags, so as to compare these before and after the model changes and fix the things which changed.


bencap (commented):
The stubs plus types look reasonable as far as I can tell. Thanks for that `dump_rels` test; really helpful + adds some good peace of mind with the SQLAlchemy upgrade. A couple minor things I was wondering about:

- Are the assertion statements generally just so that mypy understands what is possible with a variable vs. what isn't? I was wondering if we wanted some sort of assertion exception handler, but if they generally serve typing purposes for mypy that wouldn't really be needed.
- We will also need to add these new requirements to the `requirements.txt` file to make sure Docker still works. This isn't part of this work, but it might make sense to transition to something like Poetry or something similar so we can stop needing to maintain two separate dependency files (or just use pyproject.toml to install dependencies in Docker rather than the requirements file).

I think after the requirements are synchronized for Docker this looks good to me.

nickzoic:
> The stubs plus types look reasonable as far as I can tell. Thanks for that `dump_rels` test; really helpful + adds some good peace of mind with the SQLAlchemy upgrade.

Yeah, I was thinking of automating it by having an action compare the output of `tests/dump_rels.py` with a file `tests/dump_rels.txt` stored in the repo, and failing if they differ.  Then if you make any *intentional* changes to the relationships you just need to check in a new `dump_rels.txt`.   It's intentionally sorted and flat so it should diff & merge okay.

>     * Are the assertion statements generally just so that mypy understands what is possible with a variable vs. what isn't? I was wondering if we wanted some sort of assertion exception handler, but if they generally serve typing purposes for mypy that wouldn't really be needed.

Yeah, it's not the only way to do it and it's a bit ugly but it's an expedient work-around for typing things which mypy isn't *quite* smart enough to understand (and you can ask `coverage` to ignore assertion statements, so you don't get that problem).

>     * We will also need to add these new requirements to the `requirements.txt` file to make sure Docker still works. This isn't part of this work, but it might make sense to transition to something like Poetry or something similar so we can stop needing to maintain two separate dependency files (or just use pyproject.toml to install dependencies in Docker rather than the requirements file).

Oh good pick-up, I forgot about that being used: when running tests locally I have my own Dockerfile which uses `pip install -e .[dev,server]` to install all deps

(annoyingly, there's a `--no-deps` but no `--only-deps` :-) )

nickzoic:
PS: the "Removed lib/array_comparison.py" and "Removed the lib/worker.py and related stuff." commits should really be their own pull request but ended up bundled in here because getting types working for those files just to delete them anyway seemed like a waste of time.

src/mavedb/routers/score_sets.py
```diff
@@ -588,7 +583,9 @@ def create_variants(db, score_set: ScoreSet, variants_data: list[VariantData], b
     num_variants = len(variants_data)
     variant_urns = bulk_create_urns(num_variants, score_set, True)
     variants = (
-        Variant(urn=urn, score_set_id=score_set.id, **kwargs) for urn, kwargs in zip(variant_urns, variants_data)
+        # TODO: Is there a nicer way to handle this than passing dicts into kwargs
```
ashsny, line 586: as above

ashsny (approved):

PR #157, opened 2024-02-20
Aggregate Statistics for Score Sets, Experiments, and Targets 

bencap:
Adds a new statistics router endpoint to support UI summary features. These routes calculate simply counting statistics for various properties on Target, Score set, and Experiment models aggregated by property value.

To support additional search queries that might be made more frequently with these changes, additional fields were also added to the free text search function on Score sets.


tests/routers/test_statistics.py
```diff
@@ -0,0 +1,434 @@
...
+
+    response = client.get("/api/v1/statistics/record/score-set/keywords")
+    assert response.status_code == 200
+    assert response.json() == {}
+
+    response = client.get("/api/v1/statistics/record/score-set/doi-identifiers")
+    assert response.status_code == 200
+    assert response.json() == {}
+
+    response = client.get("/api/v1/statistics/record/score-set/raw-read-identifiers")
```
ashsny, line 105: you could do something like iterating over a list of these strings and provide error messages along with the asserts so it's clear which one failed, i.e.  
```
for case in cases:
  response = client.get(case)
  assert response.status_code == 200, f"non-200 status code for api endpoint: {case}"
  etc.

File: tests/routers/test_statistics.py [hunk 2]
```diff
@@ -0,0 +1,434 @@
...
+    "field_value",
+    TARGET_ACCESSION_FIELDS,
+)
+def test_target_accession_statistics(client, setup_router_db, field_value, data_files):
+    """Test target accession statistics endpoint for published score sets."""
+    experiment = create_experiment(client)
+    with patch.object(cdot.hgvs.dataproviders.RESTDataProvider, "_get_transcript", return_value=TEST_CDOT_TRANSCRIPT):
+        score_set = create_acc_score_set_with_variants(client, experiment["urn"], data_files / "scores_acc.csv")
+
+    camelized_field_value = camelize(field_value.replace("-", "_"))
```
ashsny, line 121: tests are pretty lightly commented, I'd have one comment here that says, like add one score set in the database, verify that there is one score set containing the parameterized field, blah blah.

never a bad idea to over-comment tests so that when they fail, a developer who has never seen it before will more easily be able to parse out the failure

tests/routers/test_statistics.py
```diff
@@ -0,0 +1,434 @@
...
+@pytest.mark.parametrize(
+    "field_value",
+    TARGET_SEQUENCE_FIELDS,
+)
+def test_target_sequence_statistics(client, setup_router_db, field_value, data_files):
+    """Test target sequence statistics endpoint for published score sets."""
+    experiment = create_experiment(client)
+    score_set = create_seq_score_set_with_variants(client, experiment["urn"], data_files / "scores.csv")
+
+    camelized_field_value = camelize(field_value.replace("-", "_"))
```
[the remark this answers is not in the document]
ashsny: to be specific, I don't think it would be a terrible idea to rip out the last 6 lines of each of these and just swap arround sequence/accession targetSequence/targetAccession. up to you

tests/routers/test_statistics.py
```diff
@@ -0,0 +1,434 @@
...
+    publish_score_set(client, score_set["urn"])
+    response = client.get("/api/v1/statistics/target/gene/category?")
+
+    assert response.status_code == 200
+    assert TEST_MINIMAL_ACC_SCORESET["targetGenes"][0]["category"] in response.json()
+    assert response.json()[TEST_MINIMAL_ACC_SCORESET["targetGenes"][0]["category"]] == 1
+
+
+def test_target_gene_category_statistics_seq(client, setup_router_db, data_files):
+    """Test target gene category endpoint on sequence based targest for published score sets."""
```
ashsny, line 199: r/targest/target

tests/routers/test_statistics.py
```diff
@@ -0,0 +1,434 @@
...
+# NOTE: It's clearer to not parametrize target gene statistics due to non-standard behavior between target accession
+#       and target sequence subtypes.
+def test_target_gene_category_statistics_acc(client, setup_router_db, data_files):
+    """Test target gene category endpoint on accession based targest for published score sets."""
+    experiment = create_experiment(client)
+    with patch.object(cdot.hgvs.dataproviders.RESTDataProvider, "_get_transcript", return_value=TEST_CDOT_TRANSCRIPT):
+        score_set = create_acc_score_set_with_variants(client, experiment["urn"], data_files / "scores_acc.csv")
+
+    publish_score_set(client, score_set["urn"])
+    response = client.get("/api/v1/statistics/target/gene/category?")
```
ashsny, line 191: `category?` probably deserves a comment if it's intentional

tests/routers/test_statistics.py
```diff
@@ -0,0 +1,434 @@
...
+        "ensembl-identifier": {"offset": 0, "identifier": {"dbName": "Ensembl", "identifier": "ENSG00000103275"}},
+        "refseq-identifier": {"offset": 0, "identifier": {"dbName": "RefSeq", "identifier": "NM_003345"}},
+        "uniprot-identifier": {"offset": 0, "identifier": {"dbName": "UniProt", "identifier": "Q9Y617"}},
+    }
+
+    experiment = create_experiment(client)
+
+    # Update each of the score set target genes to include our test external identifiers.
+    seq_target = TEST_MINIMAL_SEQ_SCORESET["targetGenes"][0].copy()
+    seq_target.update({"externalIdentifiers": [v for v in test_mapper.values()]})
```
ashsny, line 281: can prolly extract `{"externalIdentifiers": [v for v in test_mapper.values()]}` into a local var

ashsny (commented):
awesome PR with a lot of great fixes, just a few notes

src/mavedb/routers/statistics.py
```diff
@@ -0,0 +1,364 @@
...
+        query = (
+            db.query(
+                getattr(identifier_models[field], "identifier"),
+                func.count(getattr(identifier_models[field], "identifier")),
+            )
+            .join(association_tables[field])
+            .group_by(identifier_models[field].identifier)
+            .join(TargetGene)
+            .join(ScoreSet)
+            .filter((ScoreSet.published_date.is_not(None)))
```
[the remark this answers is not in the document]
bencap: Yup, I think there is a somewhat delicate balance between handling these in a unified way and being explicit about the DB models we are operating on. The most recent commit spins out the publication date filter into a subquery each of the stats queries joins to. It was easiest to also change the query syntax to use the 2.0 style, which matches the corresponding pure SQL syntax a little more closely.

I think that makes it better duplication wise- there are still a lot of one-off joins on a per model basis, but not sure those can really be avoided.

tests/routers/test_statistics.py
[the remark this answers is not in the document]
bencap: Makes sense, I ended up pulling out a lot of the experiment / score set setup into their own fixtures and then just making a helper function to do the assert statements. They are all pretty much the same so that eliminated a lot.

ashsny (approved):

PR #162, opened 2024-03-18
Made httpx a full (not just dev) dependency.

jstone-uw:


EstelleDa:
I think httpx is necessary for taxonomy branch too.

EstelleDa (commented):
It looks great to me.

PR #163, opened 2024-03-18
Add missing indices; fix score set deletion failure.

### LINKED ISSUE #161: Deletion of unpublished score set/experiment fails sometimes.
This problem has been reported by one user. We have not been able to reproduce it yet, and perhaps it has already been addressed. This issue should remain open until we are confident that there is no remaining issue.

jstone-uw:
Fixes #161.

The missing index on mapped_variants.variant_id has caused deletion of unpublished score sets to fail by timing out. This index was especially needed because of the table size, but while adding the index, I've added sensible indices on some other foreign keys.


jstone-uw:
In production, I manually added two indices on 2022-03-04 as a temporary fix:

- `ix_mapped_variants_js` on `mapped_variants.variant_id`
- `ix_variants_js` on `variants.scoreset_id` (no practical effect, but seemed like a good idea)

These indices should be dropped before releasing this code.

bencap (approved):

PR #165, opened 2024-03-21
Async Worker and Variant Creation Job

bencap:
Adds a new Docker service running an ARQ Redis based worker. This change uses the new queue to process score set variants in an asynchronous fashion.


docker-compose-prod.yml
jstone-uw, line 1: For production, instead of adding the worker process and Redis server to the API's Docker container, I think we'll want to
- Create a separate Docker container, with its own docker-compose-worker.yml file, that just hosts the worker.
- Use a Redis service the lives outside the Docker containers, so that we can have multiple workers talking to the same Redis service. ElastiCache is the likely option here.

Because this is a big set of changes, I wouldn't be opposed to letting the worker and Redis run on the API server for the first release, then updating the infrastructure in a patch-level release afterward. But that would only be feasible if it's okay to have two separate workers, with their own Redis servers, and if terminating a Redis server won't cause us to lose any jobs. I don't think this is the case, right?

jstone-uw (changes requested):
See my comment on docker-compose-prod.yml. Otherwise the changes look good, though they're so extensive that I'm relying on manual tests more than source code inspection.

PR #166, opened 2024-03-22
Debug creating and editing score set without target sequence or accession problem

EstelleDa:
Solve [https://github.com/VariantEffect/mavedb-ui/issues/142](https://github.com/VariantEffect/mavedb-ui/issues/142) problem. 

Without length checking, original codes can't check whether item_create.target_genes is an empty array and users can create a new score set or edit a score set without target sequence or target accession.


bencap:
Thanks for this @EstelleDa!

There was actually already a (buggy) validator I had written that was supposed to cover this, but it wasn't working correctly. Because we already had a validator set up, I figured it'd just be cleaner to fix it rather than checking the length of the target gene list (although your solution also worked perfectly fine).

Since you were on holiday, I took the liberty of fixing the validator and adding a few tests to ensure it was working properly.

EstelleDa:
Thanks @bencap ! Your codes look good to me. Manually operation doesn't show me any problem and tests all passed on my side.

PR #167, opened 2024-03-25
Better ORCID authorization

jstone-uw:
Addresses https://github.com/VariantEffect/mavedb-ui/issues/135.

The code for managing ORCID login sessions has been replaced in order to address three issues:

Users report getting logged out and not being able to log in immediately.
When using multiple tabs, users must log in separately in each tab.
The login mechanism requires a popup window.
The rationale for this change is described in lib/orcid.ts:

The changes mainly reside in mavedb-ui, but because we are switching from the OIDC implicit authentication flow to three-legged OAuth, a new API endpoint is needed that exchanges codes generated at login for ID tokens.

This update should be released in conjunction with the identically-named mavedb-ui PR, https://github.com/VariantEffect/mavedb-ui/pull/145.


.gitignore
ashsny, line 1: why this change?

src/mavedb/routers/orcid.py
```diff
@@ -0,0 +1,51 @@
...
+            "client_secret": ORCID_CLIENT_SECRET,
+            "code": request.code,
+            "grant_type": "authorization_code",
+            "redirect_uri": request.redirect_uri
+        }
+        response = await client.post(url, data=data)
+        if response.status_code == 200:
+            data = response.json()
+            print(json.dumps(data))
+            token_type = data["token_type"] # Fixed. Should be "Bearer"
```
ashsny, line 38: not sure what this comment means

PR #168, opened 2024-04-02
Show A Users' Unpublished Score Sets on an Experiment Page

bencap:
Change will display a users' unpublished score sets on an experiment page, so long as they are the signed in user. Users will still only be able to see the unpublished experiment if it is their own.

Apologies this took longer than I wanted to get up, I was trying out some UI changes as well that I didn't end up liking.

Fixes https://github.com/VariantEffect/mavedb-ui/issues/133


ashsny:
looks good to me and can go in as is. separately, we may want to indicate in the UI which are unpublished to avoid a user thinking that because they are published b/c there's no difference

bencap:
Yeah, that was my motivation for looking at some UI updates as well. 

Right now I suppose the user can know, but they'd have to infer that this is the difference between a `tmp:...` urn and a `urn:...` urn. I was thinking though maybe we could add some sort of tag (https://primevue.org/tag/) to score set UI elements, especially in the tables, that would let the user more easily see at a glance what is published, processed, incomplete, etc.

PR #170, opened 2024-04-11
Debug release-2024.0.0 branch not saving scores and counts file during creating a new score set

EstelleDa:
Main branch works well, but release2024.0.0 gets this problem. 

It happens in creating and editing a score set. Both scores and counts file can't be saved.


ashsny:
this undoes moving this work to the worker, was this an attempt to fix the problem of getting release-2024.0.0 working locally?

EstelleDa:
It seems we plan to save variants in other database. On my side, variants aren't saved successfully.
Yes, these codes save variants in local database.

PR #173, opened 2024-04-15
Public data set export

jstone-uw:


src/mavedb/lib/script_environment.py
```diff
@@ -0,0 +1,33 @@
...
+    Features:
+    - Configures logging for the script.
+    - Loads the SQLAlchemy data model.
+    - Returns an SQLAlchemy database session.
+    """
+
+    logging.basicConfig()
+
+    # Un-comment this line to log all database queries:
+    logging.getLogger("__main__").setLevel(logging.INFO)
```
ashsny, line 26: is the `# Un-comment this line` comment referring to the line it immediately precedes or the next one? In which logging config is this code intended to be submitted?

src/mavedb/scripts/export_public_data.py
```diff
@@ -0,0 +1,101 @@
...
+    .options(
+        lazyload(ExperimentSet.experiments.and_(Experiment.published_date.is_not(None))).options(
+            lazyload(Experiment.score_sets.and_(ScoreSet.published_date.is_not(None)))
+        )
+    )
+    .execution_options(populate_existing=True)
+    .order_by(ExperimentSet.urn)
+)
+
+# TODO To support very large data sets, we may want to use custom code for JSON-encoding an iterator.
```
ashsny, line 69: all TODOs should reference a github issue

src/mavedb/scripts/export_public_data.py
```diff
@@ -0,0 +1,101 @@
...
+experiment_set_views = map(lambda es: ExperimentSetPublicDump.from_orm(es), experiment_sets)
+
+# Get a list of IDS of all the score sets included. (There is no flat_map function...)
+score_set_ids = flatmap(
+    lambda es: flatmap(lambda e: map(lambda ss: ss.id, e.score_sets), es.experiments), experiment_sets
+)
+
+zip_file_name = "mavedb-dump.zip"
+
+logger.info(f"Exporting public data set metadata to {zip_file_name}/main.json")
```
ashsny, line 82: was not immediately clear to me that this contained experiment set, experiment, and score set metadata. May want to add a comment to that effect for modify the logging statement to be more explicit.

src/mavedb/view_models/score_set.py
```diff
@@ -260,3 +259,17 @@ class AdminScoreSet(ScoreSet):
...
+
+    doi_identifiers: Sequence[DoiIdentifier]
+    primary_publication_identifiers: Sequence[PublicationIdentifier]
+    secondary_publication_identifiers: Sequence[PublicationIdentifier]
+    created_by: Optional[User]
+    modified_by: Optional[User]
+    target_genes: Sequence[TargetGene]
+    private: bool
+    processing_state: Optional[ProcessingState]
+    processing_errors: Optional[dict]
```
ashsny, line 275: not sure these should be included, see above comment

src/mavedb/scripts/export_public_data.py
```diff
@@ -0,0 +1,101 @@
...
+def flatmap(f: Callable[[S], Iterable[T]], items: Iterable[S]) -> Iterable[T]:
+    return chain.from_iterable(map(f, items))
+
+
+experiment_sets_query = db.scalars(
+    select(ExperimentSet)
+    .where(ExperimentSet.published_date.is_not(None))
+    .options(
+        lazyload(ExperimentSet.experiments.and_(Experiment.published_date.is_not(None))).options(
+            lazyload(Experiment.score_sets.and_(ScoreSet.published_date.is_not(None)))
```
ashsny, line 62: I see you made processing state part of the export - I think we should also not export unprocessed score sets. it might violate user expectations re: the completeness of score sets in the export (they won't have corresponding scores files, for example).

jstone-uw: My assumption is that unprocessed score sets aren't published. Is that valid? Maybe it's possible that a published score set gets updated with new data, but I don't think we plan to allow this except by administrative intervention.

I considered excluding processing state from the JSON metadata, but figured I'd leave it in for consistency with the API.

src/mavedb/view_models/score_set.py
jstone-uw: I didn't like this much either, but my considerations were:
- Consistency with the API seemed like a useful goal to avoid confusing users.
- And there might be some future use case for a complete export in the same format.

I'll defer to your judgment, though.

src/mavedb/scripts/export_public_data.py
jstone-uw: Do lines 17-18 cover this?

jstone-uw:
Changes have been incorporated to implement our review discussion, including the decision to include a CC0 license and exclude published data not licensed under CC0.

jstone-uw:
I've incorporated Alan's last comments:
- The ZIP archive's `variants` directory has been renamed `csv`.
- Count CSV files are only generated for score sets with count columns.

PR #174, opened 2024-04-15
Debug current version of score set

EstelleDa:
Current version means this score set is the newest. None of any other score set supersedes it. `replaces_id` of `score_set` only records the superseded score set ID.

Solve https://github.com/VariantEffect/mavedb-ui/issues/75 problem.


EstelleDa:
Sorry, there're two issues. One is current version in viewing score set page, the other one is not showing superseded score sets in search results. 
The above codes only fix the first one.

PR #178, opened 2024-04-16
Add missing indices; fix score set deletion failure.

ashsny:
From Jeremy's description/comments on the original PR:

Fixes https://github.com/VariantEffect/mavedb-api/issues/161.

The missing index on mapped_variants.variant_id has caused deletion of unpublished score sets to fail by timing out. This index was especially needed because of the table size, but while adding the index, I've added sensible indices on some other foreign keys.

In production, I manually added two indices on 2022-03-04 as a temporary fix:

ix_mapped_variants_js on mapped_variants.variant_id
ix_variants_js on variants.scoreset_id (no practical effect, but seemed like a good idea)
These indices should be dropped before releasing this code.


ashsny:
looks like this one needs merge conflicts resolved

ashsny:
this one appears to be languishing - should I address the merge conflicts here and merge?

PR #179, opened 2024-04-16
Better ORCID authorization

ashsny:
Jeremy's original PR description:

Addresses https://github.com/VariantEffect/mavedb-ui/issues/135.

The code for managing ORCID login sessions has been replaced in order to address three issues:

Users report getting logged out and not being able to log in immediately.
When using multiple tabs, users must log in separately in each tab.
The login mechanism requires a popup window.
The rationale for this change is described in lib/orcid.ts:

The changes mainly reside in mavedb-ui, but because we are switching from the OIDC implicit authentication flow to three-legged OAuth, a new API endpoint is needed that exchanges codes generated at login for ID tokens.

This update should be released in conjunction with the identically-named mavedb-ui PR, https://github.com/VariantEffect/mavedb-ui/pull/145.


src/mavedb/routers/orcid.py
```diff
@@ -0,0 +1,51 @@
...
+            "client_secret": ORCID_CLIENT_SECRET,
+            "code": request.code,
+            "grant_type": "authorization_code",
+            "redirect_uri": request.redirect_uri
+        }
+        response = await client.post(url, data=data)
+        if response.status_code == 200:
+            data = response.json()
+            print(json.dumps(data))
+            token_type = data["token_type"] # Fixed. Should be "Bearer"
```
ashsny, line 38: not sure what this comment means

jstone-uw: Thanks -- sorry about the obscure comment. I removed it and added a warning just in case we ever get an unexpected token type from ORCID.

src/mavedb/lib/authentication.py
```diff
@@ -86,15 +78,15 @@ def verify_jwt(token: str) -> dict:
 
 
 async def get_access_token(
-        # access_token_query: str = Security(access_token_query),
-        access_token_header: Optional[str] = Security(access_token_header),
-        access_token_cookie: Optional[str] = Security(access_token_cookie),
+    # access_token_query: str = Security(access_token_query),
```
ashsny, line 81: why this commented out line?

.gitignore
```diff
@@ -101,7 +101,6 @@ __pypackages__/
 .env
 .env.*
 !.env.template
-!.env.dev
```
ashsny, line 104: why this change? why delete .env.dev? if deleted, every person starting on the project will be tripped up by its non-existence?

jstone-uw: The problem is that it now needs to contain our ORCID secrets. I plan to share my .env.dev file, but I don't see another way around it. What do you think?

src/mavedb/lib/authentication.py
jstone-uw: We can certainly get rid of it, but the line isn't changed in the current PR -- aside from the indentation, since Black has been run inconsistently in the past.

src/mavedb/lib/authentication.py
```diff
@@ -38,7 +30,7 @@ def decode_jwt(token: str) -> dict:
     try:
         decoded_token = jwt.decode(
             token,
-            ORCID_JWT_SIGNING_PUBLIC_KEY,
+            ORCID_JWT_SIGNING_PUBLIC_KEY or "",
```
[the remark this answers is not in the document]
jstone-uw: Your way (adding the default to os.getenv) is more elegant, thanks. I don't think it needs a comment in this case, do you?

.gitignore
ashsny: what if we have a .env.dev with a blank ORCID secret? what happens after this change in a dev instance with such an .env.dev? if most of the functionality still works, I think it's preferable to maintain that

src/mavedb/lib/authentication.py
ashsny: ah, missed that

src/mavedb/lib/authentication.py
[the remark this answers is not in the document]
ashsny: nope

PR #180, opened 2024-04-16
Use Poetry for Dependency Management

ashsny:


Dockerfile
```diff
@@ -1,4 +1,37 @@
-FROM python:3.9 AS downloader
+################################
+# python-base
+# Set up shared environment variables
+################################
+FROM python:3.9 as python-base
+
+    # Poetry
+    # https://python-poetry.org/docs/configuration/#using-environment-variables
+ENV POETRY_VERSION=1.7.0 \
```
ashsny, line 9: is there a reason to have a multi-line with `\`s here as opposed to many `ENV` statements? it looks not very ~~pythonic~~ dockerrific

Dockerfile.worker
```diff
@@ -1,51 +0,0 @@
-FROM python:3.9 AS downloader
```
ashsny, line 1: does moving this into the other dockerfile change any instructions in the `mavedb-deployment` repo?

README.md
```diff
@@ -26,20 +26,25 @@ Or add `mavedb` to your Python project's dependencies.
...
+The build utility will look at `pyproject.toml` and invoke Poetry to build the distributions.
 
-The distribution can be uploaded to PyPI using [twine](https://twine.readthedocs.io/en/stable/).
+The distribution can be uploaded to PyPI using Poetry as well. Simply invoke
+
+```
+poetry publish
+```
+
+Note that publishing the package will likely require additional options. For a list of available arguments, see the [Poetry documentation](https://python-poetry.org/docs/cli/#publish).
```
ashsny, line 47: have you tried this yet? might be good to run through it once and figure out the options.

pyproject.toml
```diff
@@ -1,14 +1,17 @@
...
-requires = ["hatchling"]
-build-backend = "hatchling.build"
+requires = ["poetry-core"]
+build-backend = "poetry.core.masonry.api"
 
-[project]
+[tool.poetry]
 name = "mavedb"
-dynamic = ["version"]
+version = "2023.5.0"
```
ashsny, line 7: is there a reason we can't do `dynamic` here? in any case, you probably want `2024.0.0` here (or `.1.0` since that's the branch this is headed for). you'll probably want to edit the wiki to reflect that this is an additional place the version number must be changed when a release is cut.

Dockerfile.worker
bencap: Yeah it will, I'll open a PR over there. It will probably be useful to try that out on staging prior to merging to work out any kinks.

README.md
bencap: No, I haven't. I'm unclear on if this is something we are still doing, since the last update to the Pypi project was last year. I can check in with others about it.

pyproject.toml
bencap: I think we may be able to just place the version here in the `pyproject.toml` file instead of in the `__version__` file and only have to worry about the one here. It seems more natural to me to have the version in the project file anyways.

In any case I'll update it to reflect the intended destination.

bencap: Hmm actually I guess that would make it tough to surface the version via the API endpoint. I'll investigate whether this is possible with Poetry.

Dockerfile
bencap: This was just usually how I saw it done in most examples. Granted I don't know Docker too well-- does each `ENV` statement create one layer? If so, I could see how it would be more straightforward to the final image if we have one `ENV` statement that does all of our poetry configuration. I figured that is why we did this for multiline `RUN` statements that are related commands.

ashsny: sounds good then

pyproject.toml
bencap: Poetry itself just duplicates the version between a `__version__` file and the `pyproject.toml` file, which I know is kinda annoying. 

It seems like [the way people get around this](https://github.com/python-poetry/poetry/issues/273#issuecomment-570999678) is by using `importlib.metadata.version('package-name')` to get the version of a package. We don't currently install the project in the pip/Poetry sense, so wouldn't be able to use that method at the moment. If we wanted to install it we could, but it would be redundant with copying the source code in the Docker file, and the only reason we'd be doing it would be to get the version string. So I guess I'm inclined to just duplicate the version.

Dockerfile.worker
bencap: I actually just looked into it and now I don't think these changes will actually necessitate any changes over in the deployment repo, which is awesome. We can confirm once we deploy this to staging.

PR #181, opened 2024-04-16
Debug deprecated score sets

ashsny:


EstelleDa:
I think it covers all of superseded score set related features.

PR #185, opened 2024-04-19
Bugfix: Possible to Create Score Sets with No Targets

### LINKED ISSUE #177: Possible to create Score Sets with no targets
It is currently possible to create score sets with no targets, which doesn't really make sense in terms of what Score Sets are meant to encompass (you can't validate any scores or counts!). We should add some validation to prevent users from doing this.

bencap:
Fixes #177 

- Fixes a bug that made it possible for users to create score sets with no targets
- Fixes a bug in a score set validator that enabled users to create score sets with duplicated target sequence labels.
- Adds test coverage for the validators in the score set view model.

### DIFF
--- a/src/mavedb/view_models/score_set.py
+++ b/src/mavedb/view_models/score_set.py
@@ -106,7 +106,7 @@ class ScoreSetModify(ScoreSetBase):
     @validator("target_genes")
     def at_least_one_target_gene_exists(cls, field_value, values):
         if len(field_value) < 1:
-            raise ValidationError("Score sets should define at least one target gene.")
+            raise ValidationError("Score sets should define at least one target.")
 
         return field_value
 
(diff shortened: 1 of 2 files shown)


src/mavedb/view_models/score_set.py
```diff
@@ -102,6 +102,14 @@ def target_labels_are_unique(cls, field_value, values):
 
         return field_value
 
+    # Validate that this score set contains at least one target attached to it
+    @validator("target_genes")
+    def at_least_one_target_gene_exists(cls, field_value, values):
+        if len(field_value) < 1:
+            raise ValidationError("Score sets should define at least one target gene.")
```
ashsny, line 109: internally, this is represented as target gene, but the error message here should probably correspond to what the user sees in the interface, which is currently 'Target Sequence' and 'Genomic Coordinates'

how about using 'target' instead of 'target gene' here?

ashsny (approved):

Issue #40, opened 2021-06-02
Keyword overhaul


afrubin:
# Rationale

MaveDB currently uses a single keyword field that can be populated with user-specified keywords. However, many of the entries are blank or have keywords of little utility (e.g. target gene name). In order to improve MaveDB’s searchability and utility for modellers, we will replace the existing optional keyword system with a new system.

Instead of one keyword field with multiple keywords, each score set or experiment will have fixed keyword categories that each take a single option. Instead of free-text keywords specified by the user, the user will select from a controlled vocabulary of terms.

# Implementation

The existing keyword field will be deprecated and no longer included in the forms or views, but the keyword data will be retained (at least for now).

New keyword fields will be created (exact names and fields to be finalized later):

- Assay System
- Variant Library Creation Method
- Selection Type
- Sequencing Modality
- Variant Scoring Method

All but the last category are specific to experiments and the last is specific to score sets. However, to preserve future flexibility, these should be implemented at the DatasetModel class level rather than in the specific subclasses. 

Although the forms will only support one keyword per entry initially, these should be implemented such that each DatasetModel can have many keywords.

Keywords should be defined in an editable JSON file, similar to the way that target reference genomes work now (see the ReferenceGenome class and its usage).

afrubin:
We now have a set of keywords and categories defined by [AVE ETS](https://www.varianteffect.org/workstreams). This work is being prepared for publication so it's timely to move forward with implementing this feature.

EstelleDa:
New keywords list from https://github.com/ave-dcd/mave_vocabulary/blob/main/schema/experiment.yml

EstelleDa:
I think we can get rid of the homepage keywords section until we decide what to do with it. The old keywords should not appear in the API and UI, but we should keep them in the DB for now. - Alan's comments.

PR #187, opened 2024-04-23
Controlled Keyword

EstelleDa:
Basic functions work, but I don't understand why some tests failed. Problem is from src/mavedb/models/experiment.py. controlled_keyword_obj of _find_or_create_keyword is none and leads to some tests failed. However, it works from UI side.


EstelleDa:
API codes of https://github.com/VariantEffect/mavedb-api/issues/40

EstelleDa:
@ashsny I don't have anything left to change in back end. Feel free to review it and give me any feedbacks.

src/mavedb/lib/validation/keywords.py
```diff
@@ -16,30 +55,38 @@ def validate_keywords(keywords: list[str]):
...
+        elif ("in vitro construct library method system" in keys) or ("in vitro construct library method mechanism" in keys):
+            raise ValidationError("Endogenous Locus Library Method does not allow in vitro method system or mechanism")
+    elif "in vitro construct library method" in values:
+        if ("in vitro construct library method system" not in keys) or ("in vitro construct library method mechanism" not in keys):
+            raise ValidationError("Miss 'In Vitro Construct Library Method System' "
+                                  "or 'In Vitro Construct Library Method Mechanism' in keywords")
+        elif ("endogenous locus library method system" in keys) or ("endogenous locus library method mechanism" in keys):
+            raise ValidationError(
+                "In Vitro Construct Library Method does not allow endogenous method system or mechanism")
+    elif "other" in values:
```
jstone-dev, line 84: It seems like this will cause problems if we allow `other` as a value for any other key in the future. It seems to me that it would be better to look at particular key-value combinations rather than looking at keys and values as separate lists. Since only one value is allowed per key, one way to do this is to turn the keywords list into a dictionary.

src/mavedb/models/experiment.py
```diff
@@ -106,32 +108,71 @@ class Experiment(Base):
...
-        # if self._updated_keywords:
-        #     return self._updated_keywords
-        # else:
-        keyword_objs = self.keyword_objs or []  # getattr(self, 'keyword_objs', [])
-        return [keyword_obj.text for keyword_obj in keyword_objs if keyword_obj.text is not None]
-
-    async def set_keywords(self, db, keywords: Optional[list[str]]):
-        if keywords is None:
-            self.keyword_objs = []
+    # Original codes. Not sure this part.
```
jstone-dev, line 111: Seems okay to me. I suppose we could use map() too, but no need.

src/mavedb/routers/experiments.py
```diff
@@ -192,6 +195,26 @@ async def create_experiment(
     for publication in publication_identifiers:
         setattr(publication, "primary", publication.identifier in primary_identifiers)
 
+    # Controlled keywords currently is allowed none. Will be changed in the future.
```
jstone-dev, line 198: If we plan to change this, let's add an item to the backlog and reference it here.

src/mavedb/lib/validation/keywords.py
EstelleDa: Thanks Jeremy! This validator should only work for the first keyword 'Variant Library Creation Method' which has these three values and shall not affect other keys that users choose 'Other'. I'll modify it. I'm wondering if it's necessary to check 'endogenous locus library method system' and 'endogenous locus library method mechanism' when users select 'Endogenous locus library method.' The same applies to 'in vitro construct library method.' For example, users only choose either 'Endogenous locus library method' or 'In vitro construct library method' in 'Variant Library Creation Method', but don't want to choose anything in other keywords.

src/mavedb/lib/validation/keywords.py
```diff
@@ -1,8 +1,47 @@
...
+from mavedb.models.controlled_keyword import ControlledKeyword
+
+
+# value will not be Optional when we confirm the final controlled keyword list.
+def find_keyword(db: Session, key: str, value: Optional[str]):
+    query = db.query(ControlledKeyword).filter(ControlledKeyword.key == key).filter(ControlledKeyword.value == value)
+    controlled_keyword = query.one_or_none()
+    if controlled_keyword is None:
+        raise ValueError(f'Invalid keyword {key} or {value}')
+    return controlled_keyword
```
bencap, line 16: Perhaps we could move this into it's own `mavedb/lib/keywords.py` file, just like how experiments and score sets have a `search_{resource}` function? Stylistically I think the functions in the validation file shouldn't interact with the database at all.

tests/validation/test_keywords.py
```diff
@@ -14,6 +22,255 @@ def test_ve_invalid_keyword(self):
         with self.assertRaises(ValidationError):
             validate_keyword(555)
 
-    def test_ve_invalid_keyword_in_list(self):
+    def test_ve_invalid_empty_keyword(self):
```
bencap, line 25: these are great, thanks!

bencap (commented):
Thanks Estelle, this will be a nice change to get out. Left some comments, lmk if you have any questions and happy to discuss anything.

I'd love for there to be some additional tests as well in the `test_experiment.py` file now that we have some more constraints on keywords. It doesn't have to be too extensive, maybe just that we are able to create an experiment with a keyword and that trying to create an experiment with an invalid keyword will do what we expect.

src/mavedb/lib/validation/keywords.py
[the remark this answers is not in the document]
EstelleDa: I agree. I think it's better. I'll modify this part.

src/mavedb/lib/validation/keywords.py
```diff
@@ -1,8 +1,47 @@
...
+    values_set = set(values)
+
+    if len(keys) != len(keys_set):
+        raise ValidationError("Duplicate keys found in keywords.")
+    if len(values) != len(values_set):
+        raise ValidationError("Duplicate values found in keywords.")
+
 
-def validate_keywords(keywords: list[str]):
+def validate_keyword(keyword: str):
```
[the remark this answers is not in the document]
EstelleDa: It's used to validate single keyword. The comment is still the previous one. I'll modify it.

src/mavedb/view_models/keyword.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    @validator("key")
+    def validate_key(cls, v):
+        keywords.validate_keyword(v)
+        return v
+
+    # @validator("value")
+    # def validate_value(cls, v):
+    #     keywords.validate_keyword(v)
+    #     return v
```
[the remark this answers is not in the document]
EstelleDa: Agree. Editing keywords also need validating.

src/mavedb/models/controlled_keyword.py
```diff
@@ -0,0 +1,19 @@
...
+
+from mavedb.db.base import Base
+
+
+class ControlledKeyword(Base):
+    __tablename__ = 'controlled_keywords'
+
+    id = Column(Integer, primary_key=True, index=True)
+    key = Column(String, nullable=False)
+    value = Column(String, nullable=False)
```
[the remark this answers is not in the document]
EstelleDa: controlled_keyword is used for Keywords. They're the options for users to choose, not the users' uploading keywords. key and value form a unique pair. Only users uploaded keywords are allowed None value now. If users don't choose anything from front end, the keywords dictionary will be like {key: "AAA", value: None}. View model's value is Optional. Otherwise, it'll get an error if users don't choose anything for that keyword and that's why I comment value's validator either. They're temporary. They'll be all compulsory when we get the final list.

src/mavedb/view_models/experiment_controlled_keyword.py
```diff
@@ -0,0 +1,44 @@
...
+    description: Optional[str]
+
+    @root_validator(pre=True)
+    def validate_fields(cls, values):
+        validated_keyword = values.get("keyword")
+        validated_description = values.get("description")
+
+        # validated_keyword possible value: {'key': 'Delivery method', 'value': None}
+        if validated_keyword and validated_keyword['value']:
+            keywords.validate_description(validated_keyword['value'], validated_keyword['key'], validated_description)
```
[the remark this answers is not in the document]
EstelleDa: It validates if value is Other, description should not be none. But if value is none, then just skip it. I'll add more details in comment.

bencap (approved):
Thanks for all the changes Estelle, looks good to me. You'll just have to resolve the merge conflict before merging this in.

PR #194, opened 2024-05-03
Expand Target Label Allowed Characters

### LINKED ISSUE #190: Updating a score set fails due to an error involving target genes.
Some score sets cannot be updated without removing and re-adding their target genes.

This is the case, for instance, with a score set that includes a target gene copied from one of the existing `TEM-15 β-lactamase` genes. My unpublished score set https://mavedb.org/#/score-sets/tmp:b66936ec-979b-4187-a02c-383d86b88b98 is an example.

Updating this score set yields validation errors:

```json
{
    "detail": [
        {
            "loc": [
                "body",
                "targetGenes",
                0,
                "targetSequence",
                "label"
            ],
            "msg": "Target sequence label `TEM-15 β-lactamase` can contain only letters, numbers, and underscores.",
            "type": "assertion_error",
            "ctx": {
                "custom_loc": null,
                "triggering_exceptions": null
            }
        },
        {
            "loc": [
                "body",
                "targetGenes",
                0,
                "__root__"
            ],
            "msg": "Expected either a `target_sequence` or a `target_accession`, not neither.",
            "type": "value_error"
        }
    ]
}
```

It seems like the second error may be a result of the first, since score sets without special characters in their target names can be updated.

bencap:
Fixes #190 

When using target labels on sequences to validate score and count data, we split variants on the same line by spaces and we split the fully qualified variant on ':' to infer the label and hgvs string. See: https://github.com/VariantEffect/mavedb-api/blob/fa1ce7d0bfd57f0e178719f462273b764a467212/src/mavedb/lib/validation/dataframe.py#L421-L424

Spaces and colons are the only characters that need to be reserved from our perspective. We can raise a validation error if a user tries to include a colon in their target label and sanitize the label of any whitespace during validation.

Adds view model tests for target sequences.

### DIFF
--- a/src/mavedb/view_models/target_sequence.py
+++ b/src/mavedb/view_models/target_sequence.py
@@ -9,6 +9,10 @@ from mavedb.lib.validation.exceptions import ValidationError
 from fqfa import infer_sequence_type
 
 
+def sanitize_target_sequence_label(label: str):
+    return label.strip().replace(" ", "_")
+
+
 class TargetSequenceBase(BaseModel):
     sequence_type: str
     sequence: str
@@ -44,7 +48,10 @@ class TargetSequenceModify(TargetSequenceBase):
         if isinstance(field_value, str):
             if ":" in field_value:
                 raise ValidationError(f"Target sequence label `{field_value}` may not contain a colon.")
-        return field_value
+
+        # Sanitize the label by stripping leading/trailing whitespace and replacing any internal whitespace with
+        # underscores. Fully qualified variants should never contain whitespace.
+        return sanitize_target_sequence_label(field_value)
 
 
 class TargetSequenceCreate(TargetSequenceModify):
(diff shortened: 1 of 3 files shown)


ashsny (approved):

src/mavedb/view_models/target_sequence.py
```diff
@@ -40,14 +44,14 @@ def validate_identifier(cls, field_value, values, field, config):
...
-                raise ValidationError(
-                    f"Target sequence label `{field_value}` can contain only letters, numbers, and underscores."
-                )
-        return field_value
+            if ":" in field_value:
+                raise ValidationError(f"Target sequence label `{field_value}` may not contain a colon.")
+
+        # Sanitize the label by stripping leading/trailing whitespace and replacing any internal whitespace with
+        # underscores. Fully qualified variants should never contain whitespace.
+        return sanitize_target_sequence_label(field_value)
```
jstone-uw, line 54: This is getting called when field_value is None, leading to an unhandled exception.

PR #201, opened 2024-05-08
Ashsny/api logging

ashsny:


src/mavedb/lib/logging.py
```diff
@@ -78,11 +78,17 @@ async def logging_route_handler(request: Request) -> Response:
...
+            old_background = response.background
             task = BackgroundTask(log_info, request, response, start, end)
-            if tasks:
-                tasks.add_task(task)
-                response.background = tasks
+            if old_background:
+                if isinstance(old_background, BackgroundTasks):
+                    old_background.add_task(task)
+                    response.background = old_background
+                else:
```
jstone-uw, line 87: What if old_background is None? Should this become a third case, where we just say `response.background = task`?

ashsny: if old_background is None, the if statement at L83 should be false, then see the else at L92

jstone-uw: Yep, thanks. I didn't read everything...

PR #225, opened 2024-05-29
Data model for external links

jstone-uw:
These changes implement the data model to support display of score-set-specific links to the UCSC Genome Browser and other sites and services.

**Notes on the data model**

- One suggested approach was to add a table for external links with a many-to-one relationship to score sets. Each row would have a site name or link type (e.g. "UCSC Genome Browser") and a URL.
- On the other hand, we could just add a column like ucsc_genome_browser_url to the score sets table.

The latter option seems too specific when we know we're likely to add more links. But I'm not sure it's worth adding table relationships for this yet, so I've opted to add a JSONB column to the score sets table. This is the intended usage of the new external_links column:

```
{
  "ucscGenomeBrowser": {
    "url": "https://genome.ucsc.edu/cgi-bin/hgTracks?db=hg38&position=chr17:43051063-43051117&hubUrl=https://public.gi.ucsc.edu/~cline/mavedb/hub/hub.txt&Variant_Effect_Maps_hideKids=1&BRCA1_00000097-s-1=dense"
  }
}
```

That is, it's an object whose keys identify the target site or service. The values are objects that currently only include URLs, but we can add metadata as necessary later. On the API side, the structure above is exposed and enforced by explicit Pydantic models.

**Release notes**

After releasing this code change, populate the new column using the attached SQL file. I've opted not to add this to the code base as a data migration, since the contents are contingent and will continue to be updated. I'm inclined to treat these links as on par with other user-contributed data.


alembic/versions/ec5d2787bec9_external_links.py
```diff
@@ -0,0 +1,34 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'ec5d2787bec9'
+down_revision = '7a345f1bf9c3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("scoresets", sa.Column("external_links", postgresql.JSONB(astext_type=sa.Text()), nullable=True))
```
bencap, line 23: As an alternative to the below, I think you could do something like 
```
op.add_column("scoresets", sa.Column("external_links", postgresql.JSONB(astext_type=sa.Text()), nullable=False, server_default='{}'))
```
which will set the default of each of the rows as the column is added so you don't need to mess with the nullable property and access the DB. See https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.Column.params.server_default

src/mavedb/view_models/score_set.py
```diff
@@ -34,6 +34,10 @@
 from mavedb.view_models.variant import VariantInDbBase
 
 
+class ExternalLink(BaseModel):
+    url: Optional[str]
```
bencap, line 38: Do we want to allow these to be None? For this link it seems like it's the difference between
```
{}
```
and
```
{"ucscGenomeBrowser": {"url": None}}
```
To me, the former seems simpler if we are going to add more links since it would reduce clutter in these objects.

Lmk if you think the other way is better though, totally open to it.

bencap (commented):
This seems like the right approach to me if we are going to add other links in the future.

alembic/versions/ec5d2787bec9_external_links.py
jstone-uw: Thanks, I like that better. I wonder why it can't generate that from the SQLAlchemy model with its default, but maybe JSONB is tricky or it doesn't want to assume that we want a server-side default.

src/mavedb/view_models/score_set.py
jstone-uw: I left `url` optional in case we want to support other links that are specified differently — with an ID or some combination of parameters rather than a plain URL. I figured I'd prevent the front end from assuming there's a URL from the beginning, while avoiding breaking API changes. But I do see both sides of the argument here.

If there's no UCSC Genome Browser link, I currently expect to find an empty object (`{}`, your first option). I agree, there should be no occasion to have the second option.

bencap: Thanks, that makes sense.

alembic/versions/ec5d2787bec9_external_links.py
bencap: Yeah not sure-- the auto-generated migrations don't handle new non-nullable columns that well imo.

bencap (approved):

alembic/versions/ec5d2787bec9_external_links.py
```diff
@@ -0,0 +1,27 @@
...
+Revises: 7a345f1bf9c3
+Create Date: 2024-05-29 06:39:17.930675
+
+"""
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.orm import Session
+
+from mavedb.models.score_set import ScoreSet
```
bencap, line 13:
```suggestion
```

Issue #80, opened 2023-09-04
Improving usage statistics


afrubin:
To demonstrate the impact of MaveDB, we are often called upon to provide detailed usage statistics. However, our current ability to collect this important information is limited.

We don't want to track individual users, but we should be collecting the following aggregated information:
* Number of page views
* Number of unique users
* Number of downloads for each dataset
* Number of external API requests
* Other things?

We'll want to collect these statistics by date and by country so that we can show that we have an international user base and also produce breakdowns over various time periods.

jstone-uw:
We discussed using Google Analytics for much of this, because with little effort it will provide deeper information than we're likely to want to collect ourselves.

In conjunction with this, I'll add basic logging of API requests, which in particular will allow us to track dataset downloads that Google Analytics will miss. For each API request, we can log the authentication method; all software clients should be using access keys rather than ORCID.

PR #248, opened 2024-07-30
Improving Application Logging

bencap:
Improves application logging on the backend in support of #80:
- Adds support for canonical logs, which emit a summary log record for each logged worker and API unit of work.  
- Uses [watchtower](https://github.com/kislyuk/watchtower) to send canonical logs to Cloudwatch for aggregation and monitoring
- Maintains a logging context (through the use of [Starlette Context](https://github.com/tomwojcik/starlette-context)) throughout the request-response cycle which allow methods to decorate the canonical loggers environment 
- Adds a similar context to worker jobs, leveraging the existing arq context
- Improves the more granular log events by tapping into canonical log context and dumping it to the log as json 
- Adds functionality for interaction identifiers that can be used to track correlated requests (requires front end support to be fully functional)


jstone-dev (approved):
I didn't run into any problems, and these changes look good to me.

- My first comment is very nit-picky: would it make sense to give the logging context interface more informative function names? People might know know what kind of context `save_to_context` refers to, and `exc_info_as_dict` seems a little obscure. It's just a matter of preference, and maybe you'd rather keep these short.

- I haven't run this in an AWS environment, but on a local dev machine it produces log messages that look like JSON wrapped in a text logging message, rather than pure JSON log messages. I thought I'd double-check that this is what you want:
```
INFO:mavedb.lib.logging.canonical:{"request_ns": 1723481603839985757, "path": "/api/v1/users/me", etc.}
```

bencap:
@jstone-dev Yeah, seems reasonable to name them better. Idk why I didn't name `save_to_context` as `save_to_logging_context`. I'll rename the other to to something a little more descriptive.

In terms of the JSON logs, I think it's only critical for them to be parsable by CloudWatch. The watchtower handler will be distinct from whatever we set up for dev output (they omit the level and __name__, see [the cloudwatch output here](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/mavedb2-staging-worker-application-logs/log-events/ea50b425ef18$252F$252Fvenv$252Fbin$252Farq$252Fmavedb.lib.logging.canonical$252F1)). I think I'll try to get watchtower to add those attributes today, they seem useful. I'll look into using the [JSON formatter](https://github.com/madzak/python-json-logger) too so that what we have locally is accurate, but a note: 

The watchtower formatter and json formatter clash by default when we want to add a custom message. For instance, the json formatter lets us write a message like this: `logger.info(message="log message", extra=logging_context()`, while still outputting a JSON object that has a `message` field. Watchtower though will interpret this as us wanting to write a log like `'log message {field1: ipsum, etc.}`. CloudWatch has trouble parsing that form, but doesn't seem to have an issue as long as the full string looks like valid JSON. I'm sure I could write a custom handler for Watchtower, but never looked into it. My feeling was that we only ever consume these log events from CloudWatch and the dev logs are still readable even without the log level and path as JSON objects, so I left it as is.

jstone-dev (approved):

Issue #254, opened 2024-08-08
It's available to save an experiment even the title, short description, abstract and methods are all empty space


EstelleDa:
Reproduce steps:
1. Enter a space in title, short description, abstract and methods in creating/editing experiment page.
2. Click the save button.

PR #255, opened 2024-08-08
Debug #254 problem

EstelleDa:
Add validators of title, short description, abstract and methods in experiment.

### DIFF
--- a/src/mavedb/view_models/experiment.py
+++ b/src/mavedb/view_models/experiment.py
@@ -74,7 +74,7 @@ class ExperimentModify(ExperimentBase):
         return v
 
     @validator("title", "short_description", "abstract_text", "method_text")
-    def validate_field_is_non_empty(cls, v, field):
+    def validate_field_is_non_empty(cls, v):
         if is_null(v) or not isinstance(v, str):
             raise ValidationError("This field is required and cannot be empty.")
         return v.strip()
--- a/src/mavedb/view_models/experiment.py
+++ b/src/mavedb/view_models/experiment.py
@@ -73,29 +73,11 @@ class ExperimentModify(ExperimentBase):
         keywords.validate_keywords(v)
         return v
 
-    @validator("title")
-    def validate_title(cls, v):
+    @validator("title", "short_description", "abstract_text", "method_text")
+    def validate_field_is_non_empty(cls, v, field):
         if is_null(v) or not isinstance(v, str):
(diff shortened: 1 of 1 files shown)


src/mavedb/view_models/experiment.py
```diff
@@ -72,6 +73,30 @@ def validate_keywords(cls, v):
...
+        if is_null(v) or not isinstance(v, str):
+            raise ValidationError("Invalid abstract text. Abstract text should not be None or space.")
+        return v
+
+    @validator("method_text")
+    def validate_method_text(cls, v):
+        if is_null(v) or not isinstance(v, str):
+            raise ValidationError("Invalid method text. Method text should not be None or space.")
+        return v
+
```
bencap, line 99:
```suggestion
    @validator("title", "short_description", "abstract_text", "method_text")
    def validate_field_is_non_empty(cls, v, field):
        if is_null(v) or not isinstance(v, str):
            raise ValidationError("This field is required and cannot be empty.")
        return v.strip()
```

bencap (changes requested):
Thanks for this Estelle. I suggested a change that will let us re-use the validator for multiple different fields, just to reduce code duplication a bit. I think we should also add a `v.strip()` to the end just to remove any leading or trailing whitespace from the field, which we also probably don't want.

src/mavedb/view_models/experiment.py
EstelleDa: Thanks Ben! I thought `is_null` function removes the space but in fact, it doesn't. I'll modify the codes.

src/mavedb/view_models/experiment.py
```diff
@@ -72,6 +73,12 @@ def validate_keywords(cls, v):
         keywords.validate_keywords(v)
         return v
 
+    @validator("title", "short_description", "abstract_text", "method_text")
+    def validate_field_is_non_empty(cls, v, field):
```
bencap, line 77:
```suggestion
    def validate_field_is_non_empty(cls, v):
```
Sorry, made a mistake in the suggestion earlier-- I don't think we need this third argument now since we don't use it anywhere.

EstelleDa: No worries. I tried `field.name` but the underscore looks weird in error messages. I agree to remove it since we don't use it anywhere.

bencap (approved):

PR #264, opened 2024-08-15
Fixed: Executing Jobs Were Sharing DB Sessions

### LINKED ISSUE #260: Issue Distinct Sessions to Worker Jobs
Currently we create a session on worker startup and give worker jobs access to it via the ARQ execution context. However, jobs using the same session might create unexpected commits or issue rollbacks at inopportune times and make other jobs non-independent. Worker jobs should be 100% independent from each other.

bencap:
Fixes #260.

Since the SQLAlchemy Session object was defined on worker startup, jobs running at the same time were sharing a worker session. This could cause unexpected commits and/or rollbacks, in addition to blocking work in the event that a job left the session in an unrecoverable state. This change creates a session at the start of each job, defining it on the job's context rather than the worker's context. This will ensure interactions between jobs and the database are fully independent.

### DIFF
--- a/src/mavedb/worker/settings.py
+++ b/src/mavedb/worker/settings.py
@@ -21,14 +21,23 @@ RedisWorkerSettings = RedisSettings(host=REDIS_IP, port=REDIS_PORT, ssl=REDIS_SS
 
 
 async def startup(ctx):
+    pass
+
+
+async def shutdown(ctx):
+    pass
+
+
+async def on_job_start(ctx):
     db = SessionLocal()
     db.current_user_id = None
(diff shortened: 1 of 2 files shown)


jstone-dev (commented):
The code changes look good and make sense to me. Should I try any manual tests? I'm guessing you've already tested worker jobs more intensively than I'm likely to, anyway.

bencap:
@jstone-dev I did some myself, if you're comfortable with that. Basically just involved trying to corrupt the worker's DB session somehow and then seeing if the session was recovered for the next job.

PR #278, opened 2024-08-29
Debug statistics record score-set and keywords problem.

EstelleDa:
Model options are from Class RecordNames(experiment and score-set) and fields options are from Class RecordFields(publication-identifiers, keywords, doi-identifiers, raw-read-identifiers and created-by). 

However, score-set doesn't have keywords anymore. Hence, adding a validator to handle this case.

### DIFF
--- a/src/mavedb/routers/statistics.py
+++ b/src/mavedb/routers/statistics.py
@@ -260,7 +260,6 @@ def _record_from_field_and_model(
     model_created_by_field = getattr(queried_model, "created_by_id")
     model_published_data_field = getattr(queried_model, "published_date")
     if field is RecordFields.createdBy:
-        print("111")
         query = (
             select(User.username, func.count(User.id))
             .join(queried_model, model_created_by_field == User.id)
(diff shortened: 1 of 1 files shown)


src/mavedb/routers/statistics.py
```diff
@@ -260,6 +260,7 @@ def _record_from_field_and_model(
     model_created_by_field = getattr(queried_model, "created_by_id")
     model_published_data_field = getattr(queried_model, "published_date")
     if field is RecordFields.createdBy:
+        print("111")
```
bencap, line 263:
```suggestion
```

src/mavedb/routers/statistics.py
```diff
@@ -269,6 +270,7 @@ def _record_from_field_and_model(
 
         return db.execute(query).all()
     else:
+        print("2222")
```
bencap, line 273:
```suggestion
```

bencap (approved):
Thanks for keeping an eye on the alerts channel Estelle, I appreciate it. If you can just remove the print statements before merging this looks good.

src/mavedb/routers/statistics.py
EstelleDa: oops. Totally forgot them. Thanks Ben!

PR #279, opened 2024-08-30
Add some basic tests.

EstelleDa:
Mainly in experiment and score set view models.


tests/routers/test_experiments.py
```diff
@@ -119,6 +119,36 @@ def test_can_update_own_private_experiment_set(session, client, setup_router_db)
...
+def test_can_update_own_private_experiment_method(session, client, setup_router_db):
+    experiment = create_experiment(client)
+    experiment_post_payload = deepcopy(TEST_MINIMAL_EXPERIMENT)
+    experiment_post_payload.update({"experimentSetUrn": experiment["experimentSetUrn"], "methodText": "New method"})
+    response = client.post("/api/v1/experiments/", json=experiment_post_payload)
+    assert response.status_code == 200
+    assert response.json()["experimentSetUrn"] == experiment["experimentSetUrn"]
+    assert response.json()["methodText"] == "New method"
+
+
```
bencap, line 151: I believe we have tests already for these operations: https://github.com/VariantEffect/mavedb-api/blob/d87b726a9ce1fbaa34cebf2ab2db1e8c5cd37501/tests/routers/test_experiments.py#L566-L583

bencap (approved):
Thanks for these Estelle! I think we already have tests for what you added in the experiment router file (linked in separate comment), but once you go ahead and remove that duplication this looks good to merge in.

tests/routers/test_experiments.py
EstelleDa: Thanks Ben! I didn't notice this test. I'll delete the duplicate tests.

PR #295, opened 2024-09-12
Score set update variant validation failures

### LINKED ISSUE #294: Editing an unpublished score set loses variants score data
Reproduce steps:
1. Click Edit button in an unpublished score set. (Any existing unpublished score set is fine. Score file doesn't matter)
2. Type anything in methods or description these kinds of fields. (NOT uploading new score file)
3. Click Save button.
4. Wait for a while, refresh the page.

View score set page will show this error message and miss variants score data.
image

bencap:
Fixes #294.

### DIFF
--- a/src/mavedb/lib/mave/utils.py
+++ b/src/mavedb/lib/mave/utils.py
@@ -1,5 +1,7 @@
 import re
 
+import pandas as pd
+
 NA_VALUE = "NA"
 
 NULL_VALUES = ("", "na", "nan", "nil", "none", "null", "n/a", "undefined", NA_VALUE)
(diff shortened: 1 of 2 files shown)


EstelleDa:
I still get the same error under this branch. It seems solve another bug.

EstelleDa:
I'm thinking whether the problem is from file src/mavedb/routers/score_sets.py line854-856, especially `.replace("NA", pd.NA)`.
```
scores_data = pd.DataFrame(
                variants_to_csv_rows(item.variants, columns=score_columns, dtype="score_data")
            ).replace("NA", pd.NA)
```

EstelleDa:
@bencap Rebuilding Docker server does solve the problem.

EstelleDa (approved):
The codes look good to me. Manually tests got a problem before but rebuilt Docker server solves it. Everything runs well on my site.

PR #306, opened 2024-09-20
Permissions for Sub-Resources

bencap:
Explicitly reference sub-resource permissions for experiment sets and experiments. These permissions were previously inferred, but with contributor additions and more complexity in permission structures planned some of these assumptions are no longer valid.


jstone-dev:
Looks good to me! I just ran a couple of simple manual tests.

jstone-dev (approved):

PR #308, opened 2024-09-30
Use Unchanging Identifiers for Worker Processes

bencap:
Previously, worker jobs used score set URNs to identify resources on which they should operate. Due to the nature of these async jobs and the fact that URNs can change on score set publication, this was a poor choice for processes that might defer execution until a moment in time after the resource identifier had changed.

This commit updates worker processes to operate on internal database identifiers to identify the resources on which they should work. This should protect us from instances where the URN changes and the worker process no longer can access the resource on which it was spawned.

### DIFF
--- a/src/mavedb/routers/score_sets.py
+++ b/src/mavedb/routers/score_sets.py
@@ -632,7 +632,7 @@ async def upload_score_set_variant_data(
         job = await worker.enqueue_job(
             "create_variants_for_score_set",
             correlation_id_for_context(),
-            item.urn,
+            item.id,
             user_data.user.id,
             scores_df,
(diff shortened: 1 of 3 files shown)


sallybg (approved):
This looks great! I like that this removes the score set argument to the variant mapper manager, it's much cleaner this way.
We talked about this offline but putting it here for visibility: If the score set is published after the mapping job starts, the job will fail because the mapper returns mapped variants associated with variant urns, and changing the score set urn also changes the variant urns. It's fine that this scenario causes the mapping to fail; the mapping job will be retried, so it will download the new variant urns upon retry and the job should then succeed.

PR #325, opened 2024-10-02
Add router functions to check whether users have authorization in experiment, experiment set and score set.

EstelleDa:
Back end codes of https://github.com/VariantEffect/mavedb-ui/issues/265
Also add some related tests.


src/mavedb/routers/experiment_sets.py
```diff
@@ -23,6 +25,40 @@
...
+
+    if user_data is not None:
+        query = query.filter(
+            or_(
+                ExperimentSet.created_by_id == user_data.user.id,
+                ExperimentSet.contributors.any(Contributor.orcid_id == user_data.user.username),
+            )
+        )
+    else:
+        return False
```
bencap, line 52: Here I think it makes more sense to use the `has_permission` function, since then we can maintain all of our permission code in a single place and make it easier to maintain.

```
access_permitted = has_permission(user_data, item, Action.XXX)  
```

src/mavedb/routers/experiment_sets.py
```diff
@@ -23,6 +25,40 @@
...
+    "/check-authorizations/{urn}",
+    status_code=200,
+    response_model=bool
+)
+async def check_experiment_set_authorization(
+    *,
+    urn: str,
+    db: Session = Depends(deps.get_db),
+    user_data: UserData = Depends(get_current_user),
+) -> bool:
```
bencap, line 38: Something I think would work really well here is leveraging the ability of Pydantic to parse enum values from a request (see: https://fastapi.tiangolo.com/tutorial/path-params/#predefined-values) and our existing `Action` enum from the permissions module. That will allow us to check the authorizations on a given resource at the given action level and make this request a little more generic without adding multiple functions.

The only thing is we need to change the way the Action enum is defined:
```
class Action(Enum):
    READ = "read"
    UPDATE = "update"
    DELETE = "delete"
    ADD_EXPERIMENT = "add_experiment"
    ADD_SCORE_SET = "add_score_set"
    SET_SCORES = "set_scores"
    ADD_ROLE = "add_role"
    PUBLISH = "publish"
```

```suggestion
@router.get(
    "/user-is-authorized/{urn}/{action}",
    status_code=200,
    response_model=bool
)
async def check_experiment_set_authorization(
    *,
    urn: str,
    action: Action,
    db: Session = Depends(deps.get_db),
    user_data: UserData = Depends(get_current_user),
) -> bool:
```

With this definition, we'll be able to make distinct requests like `/user-is-authorized/urn:00000-1/add_experiment` to check if the user can add an experiment and `/user-is-authorized/urn:00000-1/publish` to check if the user can publish. This might not be explicitly needed for this feature, but in my mind it's a nice generic way to write this endpoint that would probably end up being useful in the future.

src/mavedb/routers/experiments.py
```diff
@@ -45,6 +45,41 @@
 )
 
 
+@router.get(
```
bencap, line 48: See above comment on experiment-set

src/mavedb/routers/score_sets.py
```diff
@@ -113,6 +113,40 @@ async def fetch_score_set_by_urn(
 )
 
 
+@router.get(
```
bencap, line 116: See above comment on experiment-set

tests/routers/test_experiment_set.py
```diff
@@ -0,0 +1,45 @@
...
+
+def test_get_false_authorization_from_other_users_experiment_set_check(session, client, setup_router_db):
+    experiment = create_experiment(client)
+    change_ownership(session, experiment["urn"], ExperimentDbModel)
+    change_ownership(session, experiment["experimentSetUrn"], ExperimentSetDbModel)
+
+    response = client.get(f"/api/v1/experiment-sets/check-authorizations/{experiment['experimentSetUrn']}")
+
+    assert response.status_code == 200
+    assert response.json() == False
```
bencap, line 45: For when we check the existence of the resource, let's just add an additional test that ensures there is a 404 error returned for a non-existent resource as well.

tests/routers/test_experiments.py
```diff
@@ -562,6 +562,41 @@ def test_admin_can_update_other_users_public_experiment_set(
...
+def test_get_false_authorization_from_other_users_experiment_check(session, client, setup_router_db):
+    experiment = create_experiment(client)
+    change_ownership(session, experiment["urn"], ExperimentDbModel)
+
+    response = client.get(f"/api/v1/experiments/check-authorizations/{experiment['urn']}")
+
+    assert response.status_code == 200
+    assert response.json() == False
+
+
```
bencap, line 599: For when we check the existence of the resource, let's just add an additional test that ensures there is a 404 error returned for a non-existent resource as well.

tests/routers/test_score_set.py
```diff
@@ -496,6 +496,44 @@ def test_admin_can_add_scores_and_counts_to_other_user_score_set(session, client
...
+    experiment = create_experiment(client)
+    score_set = create_seq_score_set(client, experiment["urn"])
+    change_ownership(session, score_set["urn"], ScoreSetDbModel)
+
+    response = client.get(f"/api/v1/score-sets/check-authorizations/{score_set['urn']}")
+
+    assert response.status_code == 200
+    assert response.json() == False
+
+
```
bencap, line 536: For when we check the existence of the resource, let's just add an additional test that ensures there is a 404 error returned for a non-existent resource as well.

bencap (changes requested):
Thanks for starting on this Estelle, left a comment about how we might use Pydantic to better integrate the authorization check with our existing permissions model and checks. I think it'll be a little easier to maintain and more generalizable to potential future features if we do it in the way suggested.

src/mavedb/routers/experiment_sets.py
```diff
@@ -23,6 +25,40 @@
...
+    """
+    Check whether users have authorizations in this experiment set.
+    """
+    query = db.query(ExperimentSet).filter(ExperimentSet.urn == urn)
```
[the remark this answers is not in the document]
EstelleDa: Thanks Ben! It's better to manage in the future.

src/mavedb/routers/experiment_sets.py
EstelleDa: Thanks Ben! You give me wonderful suggestions. I considered writing a generic function before but it seemed too complex cause I didn't know we have `has_permission` already, so I gave up. I'll write a new one based on your above suggestions.

tests/routers/test_experiment_set.py
EstelleDa: I agree. After modifying the API functions, we have some error tests for it.

src/mavedb/routers/authorization.py
```diff
@@ -0,0 +1,73 @@
...
+from mavedb.lib.authentication import get_current_user, UserData
+from mavedb.lib.permissions import has_permission, Action
+from mavedb.lib.logging import LoggedRoute
+from mavedb.lib.logging.context import logging_context, save_to_logging_context
+from mavedb.models.experiment import Experiment
+from mavedb.models.experiment_set import ExperimentSet
+from mavedb.models.score_set import ScoreSet
+
+router = APIRouter(
+    prefix="/api/v1",
```
bencap, line 18:
```suggestion
    prefix="/api/v1/permissions",
```

src/mavedb/routers/authorization.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class ModelName(str, Enum):
+    experiment = "experiment"
+    experiment_set = "experiment-set"
+    score_set = "score-set"
+
+
+@router.get(
+    "/user-is-authorized/{model_name}/{urn}/{action}",
```
bencap, line 34: Love how generic this is. This was mostly on me because of how I suggested the name of the function, but I think it'd make more sense if we put this in a file named `permissions.py` and made this route `/user-is-permitted/{model_name}/{urn}/{action}`.

 The word authorization has a distinct meaning already (sign the user in) and this is more about checking a users' permissions.

src/mavedb/routers/authorization.py
```diff
@@ -0,0 +1,73 @@
...
+        if user_data:
+            try:
+                action_enum = Action[action.upper()]
+                permission = has_permission(user_data, item, action_enum).permitted
+                return permission
+            except KeyError:
+                raise HTTPException(status_code=400, detail=f"Invalid action: {action}")
+        else:
+            logger.debug(msg="Miss user data", extra=logging_context())
+            raise HTTPException(status_code=404, detail=f"User not found")
```
bencap, line 70: The `has_permission` function should be able to handle None-type user data.
```suggestion
            try:
                action_enum = Action[action.upper()]
                permission = has_permission(user_data, item, action_enum).permitted
                return permission
            except KeyError:
                raise HTTPException(status_code=400, detail=f"Invalid action: {action}")
```

src/mavedb/routers/authorization.py
```diff
@@ -0,0 +1,73 @@
...
+@router.get(
+    "/user-is-authorized/{model_name}/{urn}/{action}",
+    status_code=200,
+    response_model=bool
+)
+async def check_authorization(
+    *,
+    model_name: str,
+    urn: str,
+    action: str,
```
bencap, line 42: Here if we define the action like
```suggestion
    action: Action,
```
plus change the definition of the Action enum in the file `lib/permissions.py` to look like
```
class Action(Enum):
    READ = 'read'
    UPDATE = 'update'
    DELETE = 'delete'
    ADD_EXPERIMENT = 'add_experiment'
    ADD_SCORE_SET = 'add_score_set'
    SET_SCORES = 'set_scores'
    ADD_ROLE = 'add_role'
    PUBLISH = 'publish'
```
Fastapi will automatically validate the API call to enforce one of those strings as the action, and raise a `422` error if the action is not one of those options. For an example of that, see this page: https://fastapi.tiangolo.com/tutorial/path-params/#predefined-values

bencap (commented):
Looks nice Estelle, just a couple comments about leveraging some of the built-in FastAPI functionality when it comes to enums and some general thoughts about how to organize this new route. Thanks for all the tests too, those look great.

src/mavedb/routers/authorization.py
EstelleDa: Your previous suggestions gave me a hint for this generic function. I love this one too. I agree permission is better.

src/mavedb/routers/permissions.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@router.get(
+    "/user-is-permitted/{model_name}/{urn}/{action}",
+    status_code=200,
+    response_model=bool
+)
+async def check_authorization(
+    *,
+    model_name: str,
```
bencap, line 40: Since you went through the trouble of creating an enum for the model name, we should probably use it here instead of this being a string
```suggestion
    model_name: ModelName,
```

EstelleDa: This one done.

EstelleDa:
@bencap Sorry the new commit looks many things. I changed all of the related tests name from authorisation to permission and add one more test.

bencap (approved):
Thanks for the changes!

PR #328, opened 2024-10-07
Tests for Mapping Job

bencap:


tests/worker/test_jobs.py
```diff
@@ -428,65 +520,421 @@ async def test_create_mapped_variants_for_scoreset_with_existing_mapped_variants
...
+    ).all()
+
+    assert (await standalone_worker_context["redis"].llen(MAPPING_QUEUE_NAME)) == 1
+    assert (await standalone_worker_context["redis"].get(MAPPING_CURRENT_ID_NAME)).decode("utf-8") == ""
+    assert not result["success"]
+    assert result["retried"]
+    assert len(mapped_variants_for_score_set) == 0
+    assert score_set.mapping_state == MappingState.queued
+    assert score_set.mapping_errors is not None
+
```
sallybg, line 679: Would it be desirable to test that the retry succeeds?

tests/worker/test_jobs.py
```diff
@@ -428,65 +520,421 @@ async def test_create_mapped_variants_for_scoreset_with_existing_mapped_variants
...
         select(MappedVariant).join(Variant).join(ScoreSetDbModel).filter(ScoreSetDbModel.urn == score_set.urn)
     ).all()
+
+    assert (await standalone_worker_context["redis"].llen(MAPPING_QUEUE_NAME)) == 1
+    assert (await standalone_worker_context["redis"].get(MAPPING_CURRENT_ID_NAME)).decode("utf-8") == ""
+    assert not result["success"]
+    assert result["retried"]
     assert len(mapped_variants_for_score_set) == 0
+    assert score_set.mapping_state == MappingState.queued
+    assert score_set.mapping_errors is not None
```
sallybg, line 820: As with the other retry test, would we want to test that the retry succeeds?

sallybg (approved):
This is great, so thorough and well thought out! Only very minor comments/questions.

src/mavedb/worker/jobs.py
```diff
@@ -447,7 +440,7 @@ async def map_variants_for_score_set(
                         )
                         db.add(mapped_variant)
 
-                    if successful_mapped_variants == 0:
+                    if successful_mapped_variants == 0 and mapped_scores:
```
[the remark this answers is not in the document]
bencap: Yeah good point, I think I was misunderstanding what was going wrong diagnosing a separate issue. We shouldn't need this change.

tests/worker/test_jobs.py
bencap: I guess my thought is that we are testing that we are in a state where the retry is ready to be executed, and that the manager is the thing which is actually executing the retry. It's a little weird here because we never actually use ARQ as a queue in the tests, we just test the jobs directly, so the interaction between jobs isn't well tested (which is what you are rightly getting at).

In that vein, I'm going to add five more tests that are more end to end than the existing ones and I think cover the possible job flows:
- Does the variant creation process run end to end successfully (create variants, enqueue a manager process, and map variants)?
- Does the variant creation process halt successfully (create variants unsuccesfully, do not enqueue a manager process)?
- Does the manager execute, queue a job, and allow it to finish successfully?
- Does the manager execute, queue a job (which we will force to fail), retry and re-queue the job, and then complete successfully on the next attempt?
- Does the manager execute, queue a job which fails multiple times as desired, and then exit its execution successfully?

These tests will test that the worker after each job execution is correct, while the new tests will test that the workflow is executing properly.

PR #333, opened 2024-10-10
Modified has_permission function so that any users can add a score set to a public experiment.

EstelleDa:
And also add some related tests.

### DIFF
--- a/src/mavedb/lib/permissions.py
+++ b/src/mavedb/lib/permissions.py
@@ -177,15 +177,14 @@ def has_permission(user_data: Optional[UserData], item: Base, action: Action) ->
             else:
                 return PermissionResponse(False)
         elif action == Action.ADD_SCORE_SET:
-            return PermissionResponse(
-                (user_may_edit or roles_permitted(active_roles, [UserRole.admin])),
-                404 if private else 403,
-                (
-                    f"experiment with URN '{item.urn}' not found"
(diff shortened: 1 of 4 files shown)


src/mavedb/lib/permissions.py
```diff
@@ -177,15 +177,14 @@ def has_permission(user_data: Optional[UserData], item: Base, action: Action) ->
...
-                    if private
-                    else f"insufficient permissions for URN '{item.urn}'"
-                ),
-            )
+            if user_may_edit or roles_permitted(active_roles, [UserRole.admin]):
+                return PermissionResponse(True)
+            elif private:
+                return PermissionResponse(False, 404, f"experiment with URN '{item.urn}' not found")
+            elif user_data is not None:
+                return PermissionResponse(True)
```
bencap, line 185: Can we just add a comment here saying something to the effect of any signed in user has permissions to add a score set to an existing experiment?

bencap (approved):
Looks great, thanks for all the tests too Estelle. Just left a minor note about a comment on the new permission block.

src/mavedb/lib/permissions.py
EstelleDa: Sure! It's a good idea. In case anyone is confused about it and adding an experiment in the future.

PR #334, opened 2024-10-11
Release 2024.4.1

### LINKED ISSUE #127: support for coverage report in github actions
coverage works fine with `pytest --cov=src tests/` but that same command doesn't work properly in github actions, always reporting 0%. (see #122).

Work out why!

### LINKED ISSUE #267: Consider Adopting Ruff as our standard Formatter/Sorter/Linter
See https://docs.astral.sh/ruff/. I have used this on other projects before, and think it would be an improvement over the current Black/isort situation, plus we would get linting out of it.

### LINKED ISSUE #271: Score range data model
Data model for score ranges:
- What is the wt score?
- What ranges exist on these scores and how do they map?

Validation:
- Ranges should not overlap
- Other validation should be in the form of warnings, and can be handled in the UI.

### LINKED ISSUE #274: Update Project Dependencies
Dependabot messages have gotten more frequent, and there are now a few dependencies with high severity. We should refresh these in particular.

### LINKED ISSUE #280: During Experiment Creation, Multiple Contributor Rows for the Same Username (ORCID) May be Created
Like the users bug before this, I'm not 100% sure how this happens. We might make the index unique on usernames to prevent it from being possible though.

### LINKED ISSUE #298: Tests for Mapping Job
Tech debt: tests for mapping job

### LINKED ISSUE #301: Publishing a Score Set will Invalidate Mapping URN
Since the mapping job is based on the URN, if a user publishes a score set while mapping is underway the job will fail. We should consider performing mapping based on DB ID, or if we continue based on URN invalidating the URN in the queue if it changes.

### LINKED ISSUE #302: Make transcript accessions required for all endogenous edit score sets
Various downstream views will benefit from this change in validation behavior, and hgvs strings make more sense in this format anyways.

### LINKED ISSUE #305: Permissions are not considered on resources contained within the requested resource
For instance, if a user requests an experiment set to which they have access the API only considers permissions on the experiment set, not for contained experiments within it. This is true of experiment -> score set relationship as well (although it is less of an issue, since generally speaking the same users who have view access on experiments will have it on contained score sets).

### LINKED ISSUE #332: Users should be allowed to add a score set to any public experiment
Currently, `has_permission` function from lib/permissions.py doesn't allow users to add a score set in other's experiment. We also miss some related tests for it.

bencap:
IMPORTANT: The `LOG_CONFIG` environment variable _must_ changed to `server` upon release.

## Features/Enhancements
- #303: Closes #302 
- #287: Closes #271
- #333: For VariantEffect/mavedb-ui#262, Closes #332 
 
## Bug Fixes
- #306: Closes #305 
- #308: Closes #301 
- #285: Closes #280 

## Housekeeping
- #328: Closes #298
- #327: Closes #267
- #310: Closes #127, Closes #274


bencap:
https://github.com/VariantEffect/mavedb-ui/pull/298

jstone-dev (approved):

PR #340, opened 2024-10-18
List experiments in urn order.

EstelleDa:
For https://github.com/VariantEffect/mavedb-ui/issues/295. 

I think this is the easiest way to do it. I don't see any other place need the ordering experiments function.

### DIFF
--- a/src/mavedb/lib/experiments.py
+++ b/src/mavedb/lib/experiments.py
@@ -99,7 +99,7 @@ def search_experiments(
             )
         )
 
-    items: list[Experiment] = query.order_by(Experiment.urn).all()
+    items: list[Experiment] = query.order_by(Experiment.urn, Experiment.title).all()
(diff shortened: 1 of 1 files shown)


src/mavedb/lib/experiments.py
```diff
@@ -99,7 +99,7 @@ def search_experiments(
             )
         )
 
-    items: list[Experiment] = query.order_by(Experiment.title).all()
+    items: list[Experiment] = query.order_by(Experiment.urn).all()
```
bencap, line 102:
```suggestion
    items: list[Experiment] = query.order_by(Experiment.urn, Experiment.title).all()
```

bencap (changes requested):
Agreed this is the best spot for this. Let's still retain the title sorting, and just add the URN sort as the first order by operation.

PR #341, opened 2024-10-21
User-specific target gene search

jstone-dev:
**Notes**
- I’ve retained the general target gene search, although we don’t currently have a use case for it.
- I added simple test cases for user-specific target gene search. We could probably add more target-gene-related cases, but these cover the changes made here.
- I decided to place the user-specific target gene search in a new endpoint, /me/target-genes/search, on the model of /me/score-sets/search. An alternative would be to include a "limit to current user" parameter in the search request itself. Since we don't need users to be able to search each others' data, I thought presenting this as a separate REST collection made sense. But I'm open to changing it if someone thinks otherwise.

### DIFF
--- a/src/mavedb/lib/target_genes.py
+++ b/src/mavedb/lib/target_genes.py
@@ -1,0 +1,48 @@
+import logging
+from typing import Optional
+
+from sqlalchemy import func, or_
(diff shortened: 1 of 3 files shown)


bencap (approved):
Thanks Jeremy, looks good

PR #342, opened 2024-10-21
Add Tests for UTF-8 Encoded Score and Count Files

bencap:


jstone-dev (approved):
Since the tests pass, all I've done is read the code and check the character encodings. But it looks fine to me!

PR #344, opened 2024-10-22
Update Permitted Licenses

bencap:
Changes include:
- A new `active` property within the database that indicates whether a license is active
- The associated DB migration
- A new `active` property on License view models to persist this property in API responses
- Tests for the license router
- Tests for (1) creating score sets with an invalid license (2) modifying a score set to have an invalid license, and (3) modifying a score set which has an in-active license is possible


jstone-dev (approved):
Is there any need to allow a superseding score set to retain the deprecated license of the score set it supersedes / corrects?

Beside that, no comments; this looks good to me.

(I also wonder whether any restrictions need to be placed on licenses for meta-analyses, but the question is independent of the current changes.)

bencap:
> Is there any need to allow a superseding score set to retain the deprecated license of the score set it supersedes / corrects?
> 
> Beside that, no comments; this looks good to me.
> 
> (I also wonder whether any restrictions need to be placed on licenses for meta-analyses, but the question is independent of the current changes.)

@jstone-dev I just checked the database and the current score sets with a superseding score set are all licensed by licenses which will not be deprecated. So I'm not too worried about current score sets. I'll add a backlog item for this so it doesn't disappear as an issue though.

PR #347, opened 2024-10-24
Not showing meta-analysis experiments in creating score set dropdown menu.

EstelleDa:
API codes for https://github.com/VariantEffect/mavedb-ui/issues/296


src/mavedb/lib/experiments.py
```diff
@@ -99,6 +99,17 @@ def search_experiments(
...
+    if search.meta_analysis is not None:
+        if not search.meta_analysis:
+            query = query.filter(
+                or_(
+                    # Keep experiments without any score sets
+                    not_(Experiment.score_sets.any()),
+                    # Keep experiments where score sets exist but have no meta_analyzes_score_sets
+                    Experiment.score_sets.any(not_(ScoreSet.meta_analyzes_score_sets.any()))
+                )
+            )
```
bencap, line 111: I think we should also support the inverse logic here as well: If a user supplies `meta_analysis: true` we should return only meta analysis score sets.

src/mavedb/routers/experiments.py
```diff
@@ -43,45 +43,6 @@
 )
 
 
-# TODO: Rewrite this function.
```
bencap, line 46: Lets retain this function-- I think it still has utility for just listing all experiments.

bencap (changes requested):
Thanks for the changes Estelle! Would you also be able to add a couple tests for the newly searchable meta analysis field? It might require a new meta analysis score set to be added to the test DB.

src/mavedb/lib/experiments.py
EstelleDa: I agree. I'll add it.

src/mavedb/routers/experiments.py
EstelleDa: Sure. I'll modify it so that we don't need to send the weird q to it.

bencap: I think it is ok to just leave it as is for this PR so we can get it released. We can just handle it at some point later-- it isn't hurting anything right now.

bencap (approved):
Sorry for the delay on this Estelle, looks good!

PR #348, opened 2024-10-24
Not allowing adding score set to meta-analysis experiment.

EstelleDa:
And also front end won't show Add Score Set button in meta-analysis experiment page.

### DIFF
--- a/src/mavedb/lib/permissions.py
+++ b/src/mavedb/lib/permissions.py
@@ -180,7 +180,7 @@ def has_permission(user_data: Optional[UserData], item: Base, action: Action) ->
         elif action == Action.ADD_SCORE_SET:
             # Not allow add score set in meta-analysis experiments.
             if any(score_set.meta_analyzes_score_sets for score_set in item.score_sets):
-                return PermissionResponse(False)
+                return PermissionResponse(False, 403, f"not allow to add score set in meta-analysis experiment")
             # Only permitted users can add a score set to a private experiment.
             elif user_may_edit or roles_permitted(active_roles, [UserRole.admin]):
                 return PermissionResponse(True)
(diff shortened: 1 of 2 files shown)


src/mavedb/lib/permissions.py
```diff
@@ -178,8 +178,11 @@ def has_permission(user_data: Optional[UserData], item: Base, action: Action) ->
             else:
                 return PermissionResponse(False)
         elif action == Action.ADD_SCORE_SET:
+            # Not allow add score set in meta-analysis experiments.
+            if any(score_set.meta_analyzes_score_sets for score_set in item.score_sets):
+                return PermissionResponse(False, 403, "not allow to add score set in meta-analysis experiment")
             # Only permitted users can add a score set to a private experiment.
-            if user_may_edit or roles_permitted(active_roles, [UserRole.admin]):
+            elif user_may_edit or roles_permitted(active_roles, [UserRole.admin]):
```
bencap, line 185: Instead of this being a check in the permissions function, I think it would be better if we check this right before adding the score set. I guess I would argue the user does technically have permission to add a score set to an experiment, it is just that in this particular case we don't allow that for these particular score set types.

EstelleDa: I struggled with where to put these checking codes before. I moved it to the `create_score_set` function. I think it's better than permissions function. The final effects are the same on the MaveDB website.

src/mavedb/routers/score_sets.py
```diff
@@ -334,6 +334,10 @@ async def create_score_set(
                 msg="Failed to create score set; The requested experiment does not exist.", extra=logging_context()
             )
             raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="Unknown experiment")
+        # Not allow add score set in meta-analysis experiments.
+        if any(s.meta_analyzes_score_sets for s in experiment.score_sets):
+            raise HTTPException(status_code=status.HTTP_403_FORBIDDEN,
+                                detail="Not allow to add score set in meta-analysis experiment")
```
bencap, line 340:
```suggestion
                                detail="Score sets may not be added to a meta-analysis experiment.")
```

bencap (changes requested):
Thanks for the changes Estelle! Just had a suggestion on the response text but looks good otherwise.

bencap (approved):

PR #352, opened 2024-11-04
Target Category Constant

bencap:


jstone-dev (approved):
Looks good. I wondered at first if we really want to use custom types in the database, but on second thought it seemed like a good idea in this case to ensure correctness.

bencap:
Yeah, I was kinda torn too. I think where I fall is that I like the enums for fields that are their own database column. It's also nice to have them to take advantage of some built in pydantic validation on enum types.

PR #361, opened 2024-11-19
Contributor Provided Score Thresholds and Calibrations

bencap:


jstone-dev (commented):
Should we somehow document or validate the expected calibration keys?

Apart from this, no comments except for the one typo.

bencap:
> Should we somehow document or validate the expected calibration keys?
> 
> Apart from this, no comments except for the one typo.

@jstone-dev Are you talking about calibration keys in the sense of `pillar_project: {calibrations}` or the set of `thresholds, evidence_strengths, positive_likelihood_ratios`? 

For the former, I guess I just don't know what those will be at the moment but it does probably make sense to have a list of them which are valid-- especially before we open up an endpoint for contributors to add calibrations (if that ever happens). For the latter, I think the pydantic model validation should be sufficient for API users at least, although we should certainly work on improving our documentation in the upcoming months.

PR #365, opened 2024-12-04
Estelle/test meta analysis

EstelleDa:
Add some more basic test cases of meta analysis. Nearly the same as before, but changing the score set's creators.


tests/routers/test_score_set.py
```diff
@@ -1256,6 +1256,119 @@ def test_multiple_score_set_meta_analysis_multiple_experiment_sets_different_sco
...
+        None,
+        data_files / "scores.csv",
+        update={"title": "Test Meta Analysis", "metaAnalyzesScoreSetUrns": [score_set_1["urn"], score_set_2["urn"]]},
+    )
+    score_set_1_refresh = (client.get(f"/api/v1/score-sets/{score_set_1['urn']}")).json()
+    assert meta_score_set["metaAnalyzesScoreSetUrns"] == sorted([score_set_1["urn"], score_set_2["urn"]])
+    assert score_set_1_refresh["metaAnalyzedByScoreSetUrns"] == [meta_score_set["urn"]]
+
+    meta_score_set = (client.post(f"/api/v1/score-sets/{meta_score_set['urn']}/publish")).json()
+    assert meta_score_set["urn"] == "urn:mavedb:00000001-0-1"
```
bencap, line 1338: In so far as we check the URNs of test objects, I think we only need to check that the URN matches the URN RegEx.

tests/routers/test_score_set.py
```diff
@@ -1256,6 +1256,119 @@ def test_multiple_score_set_meta_analysis_multiple_experiment_sets_different_sco
...
+        None,
+        data_files / "scores.csv",
+        update={"title": "Test Meta Analysis", "metaAnalyzesScoreSetUrns": [score_set_1["urn"], score_set_2["urn"]]},
+    )
+    score_set_1_refresh = (client.get(f"/api/v1/score-sets/{score_set_1['urn']}")).json()
+    assert meta_score_set["metaAnalyzesScoreSetUrns"] == sorted([score_set_1["urn"], score_set_2["urn"]])
+    assert score_set_1_refresh["metaAnalyzedByScoreSetUrns"] == [meta_score_set["urn"]]
+
+    meta_score_set = (client.post(f"/api/v1/score-sets/{meta_score_set['urn']}/publish")).json()
+    assert meta_score_set["urn"] == "urn:mavedb:00000003-0-1"
```
bencap, line 1370: Same RegEx comment as above.

bencap (approved):
Looks good, thanks for adding these Estelle. Just left a couple minor comments.

tests/routers/test_score_set.py
EstelleDa: I added the RegEx checks to all meta-analysis tests, but I support keeping them cause the meta-analysis urn is more special. This way is more accurate. RegEx only confirms they match the basic rules, but can't make sure the middle one is number 0 or letters.

PR #374, opened 2025-01-02
Data collections

jstone-dev:
Points to note during review:

- The collection URN format is `urn:mavedb:collection-{uuid}`. Originally it was just `urn:mavedb:{uuid}`, but in case we have other UUID-based URNs in the future, it seemed best to include the entity type now. (UUIDs are unique, but we don't want to have to look a URN up to determine its resource type.) A format like `urn:mavedb:collections/{uuid}` also suggested itself, but this causes problems when used as a path element. The format is otherwise easy to change if you have any better idea.

- In the unit tests, there's a new facility for impersonating EXTRA_USER.

- Unit test coverage is pretty good but a bit lower in the collections router than other routers.


bencap (commented):
Looks really solid overall, left some comments about succinctness and tried to answer the TODO comments. 

Since this will be a part of a release that will be upgrading Pydantic, we might also want to think about making the GetterDict changes-- see https://github.com/VariantEffect/mavedb-api/blob/deb09a1243e5bf2e3f6c5900f8bb35d22fc2ee2a/src/mavedb/lib/validation/transform.py. I could also do that while rebasing though.

tests/routers/test_collections.py
```diff
@@ -0,0 +1,538 @@
...
+
+
+@pytest.mark.parametrize(
+     "role",
+     ContributionRole._member_names_
+)
+def test_add_collection_user_to_collection_role(role, client, setup_router_db):
+    collection = create_collection(client, {"private": True})
+
+    response = client.post(f"/api/v1/collections/{collection['urn']}/admins", json={"orcid_id": EXTRA_USER["username"]})
```
bencap, line 60:
```suggestion
    response = client.post(f"/api/v1/collections/{collection['urn']}/{role}s", json={"orcid_id": EXTRA_USER["username"]})
```

tests/routers/test_collections.py
```diff
@@ -0,0 +1,538 @@
...
+                    "firstName": TEST_USER["first_name"],
+                    "lastName": TEST_USER["last_name"],
+                    "orcidId": TEST_USER["username"],
+                },
+                {
+                    "firstName": EXTRA_USER["first_name"],
+                    "lastName": EXTRA_USER["last_name"],
+                    "orcidId": EXTRA_USER["username"],
+                },
+            ],
```
bencap, line 80:
```suggestion
            f"{role}s": expected_response[f"{role}s"].extend(
                   {
                    "firstName": EXTRA_USER["first_name"],
                    "lastName": EXTRA_USER["last_name"],
                    "orcidId": EXTRA_USER["username"],
                }
          )
        }
```

bencap (approved):
Thanks for the changes, looks good besides the two suggested changes. I don't think that test is working quite as intended at the moment.

PR #376, opened 2025-01-15
Mapped Variant Data Export

bencap:
A script for exporting all current mapped variants.


bencap:
Closing this because it's been rendered obsolete by better scripts to submit and link to ClinGen services, which was the original purpose of this feature. See #435.

PR #380, opened 2025-01-27
Annotated Variant Export

bencap:
[VA-Spec](https://va-ga4gh.readthedocs.io/en/latest/introduction.html) implementation for MaveDB mapped variant objects. VA-Spec depends on Pydantic 2.0+, which depends on Python 3.11+, so this change also bumps certain core dependencies to retain compatibility.

VA-Spec objects can be exported via a new dedicated endpoint. Objects are constructed according to VA-Spec documentation and core information model. For more information about the VA-Spec core information model, see [their documentation](https://va-ga4gh.readthedocs.io/en/latest/core-information-model/index.html).

An example VA-Spec object for a variant from https://mavedb.org/score-sets/urn:mavedb:00000097-a-1 is attached: [va-spec_example_variant.json](https://github.com/user-attachments/files/18771865/va-spec_example_variant.json)


bencap:
Tests will fail until Cat-VRS update.

bencap:
We should be all set with this feature for once the VA-Spec 1.0.0 release is ready.

PR #385, opened 2025-02-10
Remove NA columns from dowloading scores and counts files.

EstelleDa:


bencap (commented):
Thanks Estelle for the changes. I like this as a solution, my only comment is lets just replace the `download` variable names with something more descriptive of what we are actually doing (dropping NA columns). So `download` would become `drop_na_columns` and `process_downloadable_data` would become something like `drop_na_columns_from_csv_rows`. This way in the future if we want to add other custom and controllable logic we can just add another descriptive flag like this one. 

Once that is changed, this looks good to go. Let's just merge it into 2025.1.0 since this branch is on top of the branch with calibration thresholds.

EstelleDa:
> Thanks Estelle for the changes. I like this as a solution, my only comment is lets just replace the `download` variable names with something more descriptive of what we are actually doing (dropping NA columns). So `download` would become `drop_na_columns` and `process_downloadable_data` would become something like `drop_na_columns_from_csv_rows`. This way in the future if we want to add other custom and controllable logic we can just add another descriptive flag like this one.
> 
> Once that is changed, this looks good to go. Let's just merge it into 2025.1.0 since this branch is on top of the branch with calibration thresholds.
Thanks for the feedback! I think your suggestion is better. I'll modify it and merge it.

PR #387, opened 2025-02-12
Add a PubMed url test.

EstelleDa:
This is a PubMed URL test and it passes. We have direct tests already.


bencap (approved):
Thanks for investigating Estelle, looks good.

PR #396, opened 2025-03-04
Debug UnicodeDecodeError problem and add a related test.

EstelleDa:
Remove unnecessary CSV files. Regardless of whether the scores and counts files are valid, users can successfully upload them at first. The validators check the files after uploading, so the response status codes are always 200 for both creating a score set and creating variants. We already have related validators and DataFrame tests, making these CSV files unnecessary.

However, a non-UTF8 file can still cause an issue. While users may be able to open it on their computer without problems, it will raise an error when uploading to MaveDB.


bencap (approved):
Thanks for looking into those errors Estelle. Eventually, we might want to support additional data encodings but it's probably pretty low priority.

EstelleDa:
My pleasure! We could keep it as a reminder and modify it in the future.

PR #397, opened 2025-03-06
Additional Statistics Endpoints and Statistics Router Refactor

bencap:
This change adds additional statistics router endpoints:
- `/api/v1/statistics/score-set/count?group`: Count of published score sets in MaveDB, possibly grouped by month or year.
- - `/api/v1/statistics/experiment/count?group`: Count of published experiments in MaveDB, possibly grouped by month or year.
- `/api/v1/statistics/target/mapped/gene`: Counts of gene symbols to which we have mapped targets. Included in this change is a script which populates this information for historically mapped targets.
- `/api/v1/statistics/variant/count?group`: The count of variant objects in MaveDB, possibly grouped by month or year.
- `/api/v1/statistics/mapped-variant/count?group&onlyCurrent`: The count of mapped variant objects in MaveDB, possibly grouped by month or year. An additional option `onlyCurrent` controls whether to only count current mapped variants.

In addition to the new routers above, this change refactors existing routes to clarify their logic. A prior version of this router used overly complex shared functions and enum fields that shared more code at the cost of overcomplicating simple stats routes. Here, we live with slightly more code duplication rather than complex functions that are tricky to edit and maintain.

Tests for the new routes have been added as well, in addition to a new test utility that adds mapped variants to the test database.

### DIFF
--- a/src/mavedb/routers/statistics.py
+++ b/src/mavedb/routers/statistics.py
@@ -1,5 +1,7 @@
+import itertools
+from collections import OrderedDict
 from enum import Enum
-from typing import Union
+from typing import Union, Optional
 
 from fastapi import APIRouter, Depends, HTTPException
 from sqlalchemy import Table, func, select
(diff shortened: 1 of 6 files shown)


sallybg (approved):
Looks great!

PR #399, opened 2025-03-07
Fixes #358: Literal 'null' inserted for NULL score ranges.

### LINKED ISSUE #358: Literal `null` inserted for NULL Score Ranges

bencap:


sallybg (approved):
Looks great! Tests pass and distinguish between string 'null' and actual null, as expected.

PR #402, opened 2025-03-14
Materialized View Utilities for SQLAlchemy Models

bencap:
This change adds various utilities that make it easier to create and manage materialized views using our existing SQLAlchemy models.

Define a view like:
```
    class MyView(Base):
        __table__ = view(
            "my_view",
            select(
                MyModel.id.label("id"),
                MyModel.name.label("name"),
            ),
            materialized=False,
        )

    id = __table__.c.id
    name = __table__.c.name
```
and use the view as you would any other SQLAlchemy model object. We should take care to manage database migrations with views, as they are more brittle than other Alembic migrations.

As a proof of concept, this change adds a `published_variants_materialized_view` that significantly improves the performance of variant statistics endpoints.

### DIFF
--- a/alembic/versions/b85bc7b1bec7_materialized_view_for_variant_statistics.py
+++ b/alembic/versions/b85bc7b1bec7_materialized_view_for_variant_statistics.py
@@ -1,0 +1,75 @@
+"""materialized view for variant statistics
+Revision ID: b85bc7b1bec7
+Revises: c404b6719110
+Create Date: 2025-03-14 01:53:19.898198
+"""
+
+from alembic import op
+from alembic_utils.pg_materialized_view import PGMaterializedView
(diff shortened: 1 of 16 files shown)


jstone-dev (approved):
This all made sense to me. My only question is about how we'll trigger materialized view refreshes. I see the refresh utilities, but do you envision adding scripts for refreshing MVs and triggering them using some scheduler?

bencap:
@jstone-dev Yeah Nick had some examples in #343 of this sort of thing running on ARQ. We could either have a refresh job running on a schedule or we could call the refresh job as the last step of (for an MV like this one) publishing a score set. Maybe we'd even like to do both, just to make certain things are refreshed at least once per day.

I'll add it now, it should be straightforward.

PR #435, opened 2025-05-19
ClinGen & ClinVar script enhancements

jstone-dev:
Added a few options to the ClinGen submission and linkage scripts and the ClinVar control ingestion script, and fixed a few small bugs.


bencap (commented):
Thanks for these improvements Jeremy, they look good and it's great they improve the number of variants we can link. 

Two thoughts occur to me:
- We should add a couple of tests to the `test_content_constructors.py` file which test the new behavior of passing `None` to the content constructors that support it. That should probably be added to these changes.
- We should add these enhancements to the worker jobs which handle ClinGen submission and linkage. That can probably be in a new set of changes.

bencap (approved):

PR #436, opened 2025-05-22
Support mapper update for multi-target score sets

sallybg:


src/mavedb/worker/jobs.py
```diff
@@ -397,48 +404,43 @@ async def map_variants_for_score_set(
                     # TODO(VariantEffect/dcd-mapping2#3) after adding accession-based score set mapping support:
                     # this also assumes that the score set is based on a target sequence, not a target accession
```
bencap, line 405: This TODO is now addressed by these changes correct?

src/mavedb/worker/jobs.py
```diff
@@ -247,6 +248,12 @@ async def create_variants_for_score_set(
 #  Mapping variants
 ####################################################################################################
 
+ANNOTATION_LAYERS = {
+    "g": "genomic",
+    "p": "protein",
+    "c": "cdna",
+}
```
bencap, line 255: Could we move this to a new file `src/mavedb/lib/mapping.py`? Along with this layer dictionary, we can also move the VRSMap class from `src/mavedb/data_providers/services.py` to that file. That would better centralize these objects for use across the code base.

src/mavedb/worker/jobs.py
```diff
@@ -397,48 +404,43 @@ async def map_variants_for_score_set(
...
+                        excluded_pre_mapped_keys = {"sequence"}
+                        for annotation_layer in reference_metadata[target_gene_identifier]:
+                            layer_premapped = reference_metadata[target_gene_identifier][annotation_layer].get(
+                                "computed_reference_sequence"
+                            )
+                            if layer_premapped:
+                                pre_mapped_metadata[ANNOTATION_LAYERS[annotation_layer]] = {
+                                    k: layer_premapped[k]
+                                    for k in set(list(layer_premapped.keys()))
+                                    - excluded_pre_mapped_keys  # TODO does this work if no 'sequence' key?
```
bencap, line 435: Yes :)

bencap (commented):
This all looks good to me. The tests will need to be edited to reflect the new mapping results style, and we should also add more tests for accession based and multi target mappings now that they are supported.

PR #439, opened 2025-05-24
Add OddsPath Property to ScoreRange Data Model

bencap:
Adds an `OddsPath` property to the score range data model. This property is optional. If it is given, it should contain a `ratios` object with `abnormal` and `normal` (both floats), a `evidenceStrengths` property with `normal` and `abnormal` (both strings), and an optional `source` field.

Uses the PublicationIdentifierBase view model for the OddsPath source. Adds a validator to the score set record that enforces this publication identifier has been included with the score set as either a primary or secondary publication identifier. Adds tests for this check and to ensure publications are added successfully.


jstone-dev (commented):
The code looks good and has solid tests, but I have a question about the approach.

It looks like these code changes allow for two OddsPaths: normal and abnormal. But it seems to me that each score range can have an OddsPath, and I wonder if it makes more sense to associate OddsPath and evidence strength with a score range. This would give us the flexibility to allow more than two significant score ranges or only one. It also seems more general to me, and more like the way we handle calibrations in general. What do you think? I might be looking at things the wrong way.

bencap:
@jstone-dev Thanks, that's a good point. I think that makes more sense to me as well now that you point it out. Shouldn't be too hard to make that change.

src/mavedb/view_models/odds_path.py
```diff
@@ -0,0 +1,33 @@
+from typing import Literal
+from pydantic import validator
+
+from mavedb.view_models import record_type_validator, set_record_type
+from mavedb.view_models.base.base import BaseModel
+
+
+class OddsPathBase(BaseModel):
+    ratio: float
+    evidence: Literal["BS3_STRONG", "PS3_STRONG"]
```
jstone-dev, line 10: What about other evidence codes? I believe this table is complete: https://genomemedicine.biomedcentral.com/articles/10.1186/s13073-019-0690-2/tables/3

Also, I think it would be useful to allow an OddsPath without an accompanying ACMG evidence strength. I suspect it's also possible that, in the future, the two may be decoupled.

PR #447, opened 2025-05-30
Bugfix/estelle/61/num score sets and experiments

EstelleDa:
Due to the superseding score set and reading permission checks, it's a little complex than expected.


src/mavedb/routers/experiments.py
```diff
@@ -89,7 +89,13 @@ def search_experiments(search: ExperimentsSearch, db: Session = Depends(deps.get
...
     Search experiments.
     """
-    return _search_experiments(db, None, search)
+    items = _search_experiments(db, None, search)
+    if items:
+        items = [
+            enrich_experiment_with_num_score_sets(exp, None)
+            for exp in items
+        ]
+    return items
```
bencap, line 98: `_search_experiments` always returns a list, so we can just return this list comprehension directly. In the case where items is empty, the list comprehension will be empty and we will return an empty list (there's a good case to be made we should really return a 404 in that case). 
```suggestion
    return [
            enrich_experiment_with_num_score_sets(exp, None)
            for exp in items
        ]
```

src/mavedb/routers/experiments.py
```diff
@@ -105,7 +111,13 @@ def search_my_experiments(
...
     Search experiments created by the current user..
     """
-    return _search_experiments(db, user_data.user, search)
+    items = _search_experiments(db, user_data.user, search)
+    if items:
+        items = [
+            enrich_experiment_with_num_score_sets(exp, user_data)
+            for exp in items
+        ]
+    return items
```
bencap, line 120: Same as above
```suggestion
    return [
            enrich_experiment_with_num_score_sets(exp, None)
            for exp in items
        ]
```

src/mavedb/routers/experiments.py
```diff
@@ -133,7 +145,9 @@ def fetch_experiment(
         raise HTTPException(status_code=404, detail=f"Experiment with URN {urn} not found")
 
     assert_permission(user_data, item, Action.READ)
-    return item
+    updated_experiment = enrich_experiment_with_num_score_sets(item, user_data)
+
+    return updated_experiment
```
bencap, line 150: We can return the object directly
```suggestion
    return enrich_experiment_with_num_score_sets(item, user_data)
```

src/mavedb/routers/experiments.py
```diff
@@ -429,7 +450,9 @@ async def update_experiment(
     db.refresh(item)
 
     save_to_logging_context({"updated_resource": item.urn})
-    return item
+    updated_item = enrich_experiment_with_num_score_sets(item, user_data)
+
+    return updated_item
```
bencap, line 455: Same as above
```suggestion
    return enrich_experiment_with_num_score_sets(item, user_data)
```

src/mavedb/routers/score_sets.py
```diff
@@ -203,7 +220,11 @@ async def show_score_set(
     Fetch a single score set by URN.
     """
     save_to_logging_context({"requested_resource": urn})
-    return await fetch_score_set_by_urn(db, urn, user_data, None, False)
+    item = await fetch_score_set_by_urn(db, urn, user_data, None, False)
+    enriched_experiment = enrich_experiment_with_num_score_sets(item.experiment, user_data)
+    response_item = score_set.ScoreSet.from_orm(item).copy(update={"experiment": enriched_experiment})
+
+    return response_item
```
bencap, line 227: Return object directly. There are other instances in this file where we should also do this, but for brevity I won't comment on all of them
```suggestion
    return score_set.ScoreSet.from_orm(item).copy(update={"experiment": enriched_experiment})
```

tests/routers/test_score_set.py
```diff
@@ -98,6 +98,7 @@ def test_create_minimal_score_set(client, setup_router_db):
     expected_response = update_expected_response_for_created_resources(
         deepcopy(TEST_MINIMAL_SEQ_SCORESET_RESPONSE), experiment, response_data
     )
+    expected_response["experiment"].update({"numScoreSets": 1})
```
bencap, line 101: Would it make more sense to add a property to the expected response constant in `tests.helpers.constants` with this property so that we don't have to update the expected response each time?

bencap (commented):
Thanks for working on this Estelle. The calculation logic looks good, and I left a few minor comments. I messaged you on Slack about how we might fix the failing test cases.

In the future, I want to think about how to integrate some of these calculations into our Pydantic classes. It would be nice if we could just create an Experiment view model and have these fields be calculated automatically. Because of the user permissions, I think it is a bit tricky though and I want to wait until we are on Pydantic 2.0+ at least to try it.

docker-compose-dev.yml
```diff
@@ -23,7 +23,6 @@ services:
       - "8002:8000"
```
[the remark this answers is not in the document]
EstelleDa: The merge mixed mine with the original file.

src/mavedb/routers/experiments.py
EstelleDa: Thanks Ben! It also solves the incompatible type problems.

tests/routers/test_score_set.py
EstelleDa: I considered this one before. It's a dynamic value, especially when testing the permission, whether private or public, for contributors and superseding cases. That's why I finally don't put it in `tests.helpers.constants`. Even though we put it in `tests.helpers.constants`, most of the time we still need to update it manually. I modify this part to set it to 0 in `tests.helpers.constants`, and manually update it on the rest tests. I don't have any other better solution for this part.

bencap (approved):

PR #464, opened 2025-06-21
Enhance Clingen Allele ID lookup with logging and error handling

### LINKED ISSUE #463: Return 404 for clingen allele id searches which don't yield matching MaveDB variants

bencap:
Closes #463 

- Added logging context for requested Clingen Allele IDs and found variants.
- Implemented error handling to raise HTTP 404 if no variants are found for the provided IDs.


src/mavedb/routers/variants.py
```diff
@@ -42,12 +46,24 @@ def lookup_variants(
...
+    save_to_logging_context(
+        {"clingen_allele_ids_with_permitted_variants": num_variants_matching_clingen_allele_ids_and_permitted}
+    )
+
+    if not any(matched_variants for matched_variants in variants_by_allele_id.values()):
+        logger.info(msg="No variants found for the provided Clingen Allele IDs.", extra=logging_context())
+        raise HTTPException(status_code=404, detail="No variants found for the provided Clingen Allele IDs.")
 
-    return [variants_by_allele_id[allele_id] for allele_id in request.clingen_allele_ids]
+    return list(variants_by_allele_id.values())
```
jstone-dev, line 66: Would it be worthwhile to add a note about reliance on Python's (3.7+) preservation of dictionary key order? Maybe it's unnecessary by now; I'd follow your judgment.

jstone-dev (approved):

src/mavedb/routers/variants.py
bencap: I guess is the client able to assume that the order of the variants we return is the same as the order in which they were queried? If so I can add a note, but I guess when I was writing this I hadn't consciously enforced the order being maintained.

PR #485, opened 2025-07-31
UniProt ID Mapping for Targets with Post Mapped Metadata

bencap:
This pull request introduces a new feature for mapping UniProt IDs from metadata and integrates it into the existing system. It includes database schema updates, new utility functions, API integration for UniProt ID mapping, and enhancements to the worker job processing pipeline.

### Database and Schema Updates:
* Added a new column `uniprot_id_from_mapped_metadata` to the `target_genes` table to store UniProt IDs derived from metadata (`alembic/versions/4b9fb51a460d_add_uniprot_id_field_for_mapped_metadata.py`, `src/mavedb/models/target_gene.py`, [src/mavedb/view_models/target_gene.pyR89](diffhunk://#diff-fa19d403bcbad0924715928a7da3e6504cfb6ce4a4d13a724b522460bb0b6d04R89)).

### UniProt API Integration:
* Implemented the `UniProtIDMappingAPI` class for interacting with the UniProt REST API, including methods to submit ID mapping jobs, check job status, and retrieve results (`src/mavedb/lib/uniprot/id_mapping.py`).
* Added constants for UniProt API configuration (`src/mavedb/lib/uniprot/constants.py`).
* Created utility functions to infer database names and extract sequence accession IDs from metadata (`src/mavedb/lib/uniprot/utils.py`, `src/mavedb/lib/mapping.py`).

### Worker Job Enhancements:
* Updated `map_variants_for_score_set` to handle UniProt ID mapping jobs, including error handling and job tracking (`src/mavedb/worker/jobs.py`, [[1]](diffhunk://#diff-23dc1425b054f2fa3a79e249538653d5e66f440dfea897cc34a4c0d8d00f1e4fL308-R314) [[2]](diffhunk://#diff-23dc1425b054f2fa3a79e249538653d5e66f440dfea897cc34a4c0d8d00f1e4fL384-R394).
* Introduced new exception classes for UniProt ID mapping errors (`src/mavedb/lib/exceptions.py`).

### CLI and Environment Configuration:
* Added a CLI script to process UniProt ID mappings for score sets, with options for polling intervals and target databases (`src/mavedb/scripts/map_to_uniprot_id_from_mapped_metadata.py`).
* Updated the `.env.template` file to include a toggle for enabling UniProt ID mapping (`settings/.env.template`).

These changes collectively enable the system to map UniProt IDs from metadata, integrate with the UniProt API, and handle the associated workflows efficiently.


src/mavedb/scripts/map_to_uniprot_id_from_mapped_metadata.py
```diff
@@ -0,0 +1,97 @@
...
+    "UniRef100",
+]
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@with_database_session
+@click.option("--score-set-urn", type=str, default=None, help="Score set URN to process. If not provided, process all.")
+@click.option("--polling-interval", type=int, default=3, help="Polling interval in seconds for checking job status.")
```
davereinhart, line 29: No change required, but I was finding the default interval to be too short here, and reaching the max number of retries before the job was finished. Results may vary, but I had better success when setting this to 40.

src/mavedb/scripts/map_to_uniprot_id_from_mapped_metadata.py
```diff
@@ -0,0 +1,97 @@
...
+@click.option("--polling-attempts", type=int, default=5, help="Number of tries to poll for job completion.")
+@click.option("--to-db", type=str, default="UniProtKB", help="Target UniProt database for ID mapping.")
```
davereinhart, line 31: Maybe worth adding another option to skip target genes that already have a `uniprot_id_from_mapped_metadata` value, or make that the default behavior with an option to override?

bencap: Yup, that would be useful. I added two new flags:
- `--prefer-swiss-prot` (true by default) prefers swiss prot mapping results when available. If no swiss prot entry is available, it will return all entries. If any are available, it will return them all.
- `--refresh-mapped-identifier` (false by default) will refresh the mapped uniprot id if it exists. If this flag is off and an identifier exists, the target will be skipped.

Issue #518, opened 2025-09-23
Generic Score Ranges for Supported Formats


bencap:
At present, we accept two flavors of score ranges: Brnich style ranges with odds paths and calibration style Zeiberg ranges. For the Brnich style ranges, each range really has the same underlying data to display the range and the difference lies in the source used to generate them. This differs from the implication of how ranges are stored in MaveDB: under an explicitly named key in the range dictionary.

This isn't really a problem for now, but as we continue to develop the database we'll likely want to accept various flavors of supported range styles. This will require a bit of a rework to the Pydantic container that validates the ranges, but should be possible. Right now, each range is defined explicitly and expected under a unique key in the container:
https://github.com/VariantEffect/mavedb-api/blob/d088e1abb47e89d023f7d4abe7d3d519f9755703/src/mavedb/view_models/score_range.py#L384-L390

We might consider retaining certain reserved keys while forcing arbitrary key/value pairs in the dictionary validate their values against all possible range definitions. We'd retain the ability to easily access certain important ranges (such as investigator provided ranges) while gaining the flexibility to accept other arbitrarily named ranges and/or calibrations users may want to supply for certain data sets.

This might be possible by using the Pydantic RootModel as below, but I haven't really explored it.
```
RangeTypes = Union[SavedInvestigatorScoreRanges, SavedZeibergCalibrationScoreRanges]

class RangeModel(RootModel[Dict[str, RangeTypes]]):
    investigator_provided: Optional[SavedInvestigatorScoreRanges] = None
    zeiberg_calibration: Optional[SavedZeibergCalibrationScoreRanges] = None

range_data = {
    "investigator_provided": <Instance of SavedInvestigatorScoreRanges>,
    "arbitraryRange1": <Instance of SavedInvestigatorScoreRanges>,
    "arbitraryRange2": <Instance of SavedZeibergCalibrationScoreRanges>,
    "zeiberg_calibration": None,
}
range_instance = RangeModel(range_data)
```

We should:
- Allow additional calibrations with custom names, validated against existing calibration models.
- Consider whether the utility of retaining statically named calibrations for certain calibration types important for site function is worth it. This mostly comes down to ease of type checking and always knowing that, for instance, the investigator provided ranges are of the correct type.
    - If we don't anticipate this being a large issue, we might consider just transitioning this dictionary to a list.
- Add validation to ensure generic calibrations keys don't collide with existing names.
[...]

PR #519, opened 2025-09-24
Additional Brnich Style Calibrations

bencap:
- Rename `PillarProject` ranges to `ZeibergCalibration`
- Add `research_use_only` and `title` properties to score ranges
- Generalizes the 'Brnich' style to a separate range class (This will be useful for #518) 
- Adds Scott calibration model, another form of Brnich style ranges

### DIFF
--- a/src/mavedb/view_models/score_range.py
+++ b/src/mavedb/view_models/score_range.py
@@ -87,9 +87,9 @@ class ScoreRange(SavedScoreRange):
 
 
 class ScoreRangesBase(BaseModel):
-    title: str = "Score Ranges"
+    title: str
+    research_use_only: bool
     ranges: Sequence[ScoreRangeBase]
-    research_use_only: bool = False
     source: Optional[Sequence[PublicationIdentifierBase]] = None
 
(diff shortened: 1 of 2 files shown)


bencap:
> I'm not sure we ought to have classes dedicated to the Scott calibrations, which are just for one score set. I think the more likely scenario is that we want to accommodate multiple Brnich-style calibrations, alongside multiple other calibrations.
> 
> Would it be possible to do this by turning ScoreSetRangesBase into a dictionary like this?
> 
> ```
> class ScoreRangesBase(BaseModel):
>     data: dict[str, Calibration]
> ```
> 
> where Calibration can be either InvestigatorScoreRangesBase or ZeibergCalibrationScoreRangesBase?

Yup for sure. We'll handle this in #518

bencap:
Database keys to be changed with 

```
UPDATE
  scoresets
SET
  score_ranges = jsonb_set(score_ranges - 'pillar_project', '{zeiberg_calibration}', score_ranges -> 'pillar_project')
WHERE
  score_ranges ? 'pillar_project';
```

PR #525, opened 2025-09-30
Score set search result optimization

jstone-dev:
- Add a limit option to score set search queries. Require a limit of at most 100 for searches of all score sets, while searches for the user's own score sets can have no limit.
- Extract the score set search query filter clause logic into a new function.
- Use limit + 1 in the search query, and if limit is exceeded, run a second query to count available rows. In either case, return the number of available rows with the limited search result.
- Instead of searching all score sets and then replacing un-superseded ones with their successors, revise the database query to search only un-superseded score sets.
- In the main search endpoint (but not in the "my score sets" endpoint), mandate that the search be only for published score sets.
- Add an endpoint to obtain search filter options based on a given score set search.
- Add an endpoint to obtain search filter options based on a given score set search.

### DIFF
--- a/src/mavedb/lib/score_sets.py
+++ b/src/mavedb/lib/score_sets.py
@@ -1,8 +1,9 @@
+from collections import Counter
 import csv
 import io
 import logging
-import re
 from operator import attrgetter
+import re
 from typing import Any, BinaryIO, Iterable, Optional, TYPE_CHECKING, Sequence, Literal
 
 from mavedb.models.mapped_variant import MappedVariant
@@ -10,7 +11,7 @@ import numpy as np
 import pandas as pd
 from pandas.testing import assert_index_equal
 from sqlalchemy import Integer, and_, cast, func, or_, select
-from sqlalchemy.orm import Session, aliased, contains_eager, joinedload, selectinload
+from sqlalchemy.orm import Session, aliased, contains_eager, joinedload, Query, selectinload
 
 from mavedb.lib.exceptions import ValidationError
 from mavedb.lib.logging.context import logging_context, save_to_logging_context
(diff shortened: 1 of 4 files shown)


bencap (commented):
Thanks for writing this Jeremy, it'll be quite a nice improvement to the UX of landing on the search page. 

The only other thing I was wondering is if we might add a `page` parameter to the search as well. Given the implementation is just slicing the score set list, it seems to me like it would be really easy to add (`score_sets[limit * page : limit * page + 1]`) and would complete the feature. I'm not sure how many people would practically click on a `next` button on the UI, but it would be annoying to me if I had a search with say 150 results and I was categorically denied from viewing the final 50. It would need a few tests though too. 

I know this is meant to be more of a stopgap feature though, so if you don't think it's worth the additional effort to add the tests for it we can leave it off.

src/mavedb/lib/score_sets.py
```diff
@@ -71,11 +72,15 @@ def options(cls) -> list[str]:
...
-    save_to_logging_context({"score_set_search_criteria": search.model_dump()})
+def build_search_score_sets_query_filter(
+    db: Session, query: Query[ScoreSet], owner_or_contributor: Optional[User], search: ScoreSetsSearch
+):
+    superseding_score_set = aliased(ScoreSet)
 
-    query = db.query(ScoreSet)  # \
-    # .filter(ScoreSet.private.is_(False))
+    # Limit to unsuperseded score sets.
+    # TODO#??? Prevent unpublished superseding score sets from hiding their published precursors in search results.
```
bencap: Did we end up opening an issue for this we can add the number for?

PR #541, opened 2025-10-17
Modify the function of get_score_set_variants_csv to allow downloading multiple data types together.

EstelleDa:


src/mavedb/lib/score_sets.py
```diff
@@ -437,18 +437,33 @@ def get_score_set_variants_as_csv(
...
+            dataset_columns = [str(x) for x in list(score_set.dataset_columns.get(column, []))]
+            if column == "score_columns":
+                for c in dataset_columns:
+                    prefixed = "scores." + c
+                    columns.append(prefixed)
+            elif column == "count_columns":
+                for c in dataset_columns:
+                    prefixed = "counts." + c
+                    columns.append(prefixed)
+    elif len(data_types) == 1 and data_types[0] == "scores":
```
bencap:
```suggestion python
    namespaced_score_set_columns = {k: {} for k in data_types}

    columns = ["accession", "hgvs_nt", "hgvs_splice", "hgvs_pro"]
    if include_post_mapped_hgvs:
        columns.append("post_mapped_hgvs_g")
        columns.append("post_mapped_hgvs_p")

    if include_custom_columns:
        if "scores" in namespaced_score_set_columns:
            namespaced_score_set_columns["scores"] = {
                col for col in [str(x) for x in list(score_set.dataset_columns.get("score_columns", []))]
            }
        if "counts" in namespaced_score_set_columns:
            namespaced_score_set_columns["counts"] = {
                col for col in [str(x) for x in list(score_set.dataset_columns.get("count_columns", []))]
            }

    elif "scores" in data_types:
        columns.append(REQUIRED_SCORE_COLUMN)
```
First, we might consider maintaining a dictionary of all of our namespaces. Each namespace keys a list of column names. The idea here is that if we combine the outer key and any column name from the contained list, we get a namespaced column. This will let us both easily generate a namespaced column name while also retaining efficient and easy access to the real internal column name. 

In this suggestion, I also eliminated the dictionaries for the internal mappings of the column keys. I think that we are only ever going to have two of those, and doing each of them explicitly is more readable.

Note that on the last line, I think we want to add the score column whenever scores is in the data types list and we aren't including custom columns. Right now if you supplied `[clinvar, scores]` and `include_custom_columns = False`, you wouldn't get the score column as output which isn't what we want.

src/mavedb/lib/score_sets.py
bencap: The biggest difference in this implementation is here, where we now stop falling back on custom keys in the `else` block and instead handle all keys in each namespace explicitly. If a column name is in a namespaced columns list and it isn't found by this function, it gets a value of `None`. I like this though because we are handling exports at the namespace level and we have to explicitly define the export behavior for each key.

This is also where we rename the columns so we can get their namespaces. Since we still have access to the raw column name, it's more readable to do that renaming and we don't need to do anything with string manipulation.

(i renamed the `dtype` parameter to `namespaces`.

```suggestion
    row: dict[str, Any] = {}

    # Handle each column key explicitly as part of its namespace.
    for column_key in namespaces.get("core", []):
        if column_key == "hgvs_nt":
            value = str(variant.hgvs_nt)
        elif column_key == "hgvs_pro":
            value = str(variant.hgvs_pro)
        elif column_key == "hgvs_splice":
            value = str(variant.hgvs_splice)
        elif column_key == "accession":
            value = str(variant.urn)

        if is_null(value):
            value = na_rep
        
        # export columns in the `core` namespace without a namespace
        row[column_key] = value

    for column_key in namespaces.get("mavedb", []):
        if column_key == "post_mapped_hgvs_g":
            hgvs_str = get_hgvs_from_post_mapped(mapping.post_mapped) if mapping and mapping.post_mapped else None
            if hgvs_str is not None and is_hgvs_g(hgvs_str):
                value = hgvs_str
            else:
                value = ""
        elif column_key == "post_mapped_hgvs_p":
            hgvs_str = get_hgvs_from_post_mapped(mapping.post_mapped) if mapping and mapping.post_mapped else None
            if hgvs_str is not None and is_hgvs_p(hgvs_str):
                value = hgvs_str
            else:
                value = ""

        if is_null(value):
            value = na_rep

        row[f"mavedb.{column_key}"] = value

    for column_key in namespaces.get("scores", []):
        parent = variant.data.get("score_data") if variant.data else None
        value = str(parent.get(column_key)) if parent else na_rep
        row[f"scores.{column_key}"] = value

    for column_key in namespaces.get("counts", []):
        parent = variant.data.get("count_data") if variant.data else None
        value = str(parent.get(column_key)) if parent else na_rep
        row[f"counts.{column_key}"] = value

    return row
```

EstelleDa: Thanks for sharing the code! They look much clearer and easier to read. I can't commit them directly so I copied and modified them. I'll commit them when the tests pass.

src/mavedb/lib/score_sets.py
bencap: Yeah that's fair. We could always change the column names. If we're going to have namespaces for things based on application though like `clinvar` and `gnomad`, maybe it does make sense for it to go in a `mavedb` namespace for consistency. Feel free to leave as `mavedb` for now and we can revisit it later if necessary.

EstelleDa: I keep it as `mavedb` first. We can change it later if necessary.

src/mavedb/routers/score_sets.py
```diff
@@ -1241,24 +1254,22 @@ async def update_score_set(
...
 
             scores_data = pd.DataFrame(
-                variants_to_csv_rows(item.variants, columns=score_columns, dtype="score_data")
+                variants_to_csv_rows(item.variants, columns=score_columns)
             ).replace("NA", pd.NA)
 
             if item.dataset_columns["count_columns"]:
                 count_data = pd.DataFrame(
-                    variants_to_csv_rows(item.variants, columns=count_columns, dtype="count_data")
+                    variants_to_csv_rows(item.variants, columns=count_columns)
```
bencap: I think we will need to change how this works slightly. Probably the easiest way is to add another parameter to the `variants_to_csv_row` function `namespaced=True`. As written, the re-validated data will be stored with namespaces when we want it to be stored with just the column names.

If we add the `namespaced` parameter to the `variants_to_csv_rows` function, we can pass it `namespaced=False` in this context so that it doesn't namespace any of the column names and then doesn't save those columns with the namespace. In other contexts, we will get the namespaces back.

We'd need to add some additional logic to the `variants_to_csv_rows` to handle the new parameter, but that would be pretty straightforward.

EstelleDa: I'll modify this part next week. Branch release-2025.5.0 has some conflict parts.

PR #546, opened 2025-10-20
Adds column metadata support and a multipart PATCH endpoint to update score sets and variants

davereinhart:
**Summary**
Adds column metadata support and a multipart PATCH endpoint to update score set metadata and variant data together with conditional variant regeneration.

**Key Changes**

- New dataset_columns (score/count column names + metadata)
- New ScoreSetUpdateAllOptional model with as_form for multipart parsing
- PATCH /api/v1/score-sets-with-variants/{urn} accepting:
  - scores_file, counts_file (optional)
  - score_columns_metadata_file, count_columns_metadata_file (JSON)
  - Partial metadata fields (title, publications, target genes, scoreRanges, etc.)
- Target gene find_or_create helpers to prevent duplicate rows
- Conditional variant rebuild (only on files or target gene changes)
- Permission split: UPDATE vs SET_SCORES for file/variant changes
- Processing state set to processing before enqueue
- Worker fixtures for column metadata
- Updated tests for new models and flows

**Rationale**
Unifies metadata and data updates, reduces redundant variant rebuilds, and introduces structured column annotations for downstream processing.

**Backwards Compatibility**
PUT still works; new fields are additive. No behavior change for metadata-only updates.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 # python-base
 # Set up shared environment variables
 ################################
-FROM --platform=amd64 python:3.11 AS python-base
+FROM python:3.11 AS python-base
 
     # Poetry
     # https://python-poetry.org/docs/configuration/#using-environment-variables
--- a/src/mavedb/db/view.py
+++ b/src/mavedb/db/view.py
@@ -6,8 +6,8 @@ from functools import partial
 
 import sqlalchemy as sa
 from sqlalchemy.ext import compiler
-from sqlalchemy.schema import DDLElement, MetaData
 from sqlalchemy.orm import Session
+from sqlalchemy.schema import DDLElement, MetaData
 
 from mavedb.db.base import Base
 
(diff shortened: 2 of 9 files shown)


src/mavedb/view_models/score_set.py
```diff
@@ -375,9 +449,9 @@ def publication_identifiers_validator(cls, value: Any) -> list[PublicationIdenti
         assert isinstance(value, Collection), "Publication identifier lists must be a collection"
         return list(value)  # Re-cast into proper list-like type
 
-    @field_validator("dataset_columns")
-    def camelize_dataset_columns_keys(cls, value: dict) -> dict:
-        return camelize(value)
+    # @field_validator("dataset_columns")
+    # def camelize_dataset_columns_keys(cls, value: dict) -> dict:
+    #     return camelize(value)
```
bencap:
```suggestion
```
Can remove unused obsolete validator

src/mavedb/view_models/target_gene.py
```diff
@@ -2,7 +2,7 @@
 from typing import Any, Optional, Sequence
 from typing_extensions import Self
 
-from pydantic import Field, model_validator
+from pydantic import Field, model_validator, ConfigDict
```
bencap:
```suggestion
from pydantic import Field, model_validator
```

src/mavedb/lib/validation/dataframe/dataframe.py
```diff
@@ -162,6 +185,31 @@ def validate_dataframe(
...
+        The standardized dictionary
+    """
+
+    def clean_key(key: str) -> str:
+        key = key.strip()
+        # Only remove quotes if the key is fully quoted
+        if (key.startswith('"') and key.endswith('"')) or (key.startswith("'") and key.endswith("'")):
+            key = key[1:-1]
+
+        return key.strip()
```
bencap: Could probably extract this into another method `clean_col_name(col: str)` (or similar) and merge it with the function on lines 237-243 so that we have the same logic for the column cleaning in both spots.

tests/view_models/test_score_set.py
```diff
@@ -382,66 +383,88 @@ def test_saved_score_set_synthetic_properties():
...
+        ("license_id", EXTRA_LICENSE["id"]),
+        ("target_genes", TEST_MINIMAL_SEQ_SCORESET["targetGenes"]),
+        ("score_ranges", TEST_SCORE_SET_RANGES_ALL_SCHEMAS_PRESENT),
+    ],
+)
+def test_score_set_update_all_optional(attribute, updated_data):
+    ScoreSetUpdateAllOptional(**{attribute: updated_data})
+
+
+# def test_saved_score_set_data_set_columns_are_camelized():
```
bencap: We can remove the commented test since it's no longer applicable

bencap (commented):
Looks really nice and should be a solid base for anything else we do with column metadata. Thanks for the related changes to the enqueue duplication, even though that was ancillary that is great to get cleaned up!

Comments are pretty minor and I left a couple about refactoring out some methods / classes such that they could be re-used later.

Before merging I would love to get a couple more test cases that cover the two `find_or_create_target` functions (two each for find and create) plus two tests on the `score_set_update` function that asserts the boolean field is set correctly for a target that doesn't need an enqueue and for a target that does. After that, I think the logic all looks good to merge.

src/mavedb/view_models/target_gene.py
```diff
@@ -25,6 +25,7 @@ class TargetGeneBase(BaseModel):
     category: TargetCategory
     external_identifiers: Sequence[external_gene_identifier_offset.ExternalGeneIdentifierOffsetBase]
 
+    model_config = ConfigDict(from_attributes=True)
```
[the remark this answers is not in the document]
bencap: I think I lean towards removing them. Maybe the standard will be that if there is a PATCH endpoint the associated view model will have an update model, if not just leave it at a modify.

bencap (commented):
Looks great, thanks for the updates! Just had one comment on how we handle certain special fields in score set updates. After we address that this looks gtg.

PR #548, opened 2025-10-22
Add Gzip middleware to fastapi, and score set search parameter to optionally skip experiment score set URNs and count

davereinhart:
Improves overall performance by adding gzip middleware to fastapi to compress large responses.

Adds new optional boolean parameter to `ScoreSetsSearch` and updates score set search endpoint to make retrieval of parent experiment score set URNs and count optional.


bencap:
> @bencap Once this change is deployed, wouldn't the content-length header just show the size of the compressed response? If that's the case, this type of logging may not be as informative as if were put in place beforehand.

Hmm yeah that's true, I guess for it to work properly with these changes we'd need to inject a pre-compressed content length into the logging context via a middleware that runs prior to the gzip compression.

PR #553, opened 2025-10-29
Populate and surface post-mapped HGVS expressions and VEP functional consequence, and surface gnomAD AF

sallybg:


src/mavedb/scripts/populate_mapped_hgvs.py
```diff
@@ -0,0 +1,188 @@
...
+
+                    elif mapped_variant.clingen_allele_id.startswith("PA"):
+                        # if PA, assume that assay was performed at amino acid level, so only provide hgvs_p
+                        if data.get("aminoAcidAlleles"):
+                            for allele in data["aminoAcidAlleles"]:
+                                if allele.get("hgvs"):
+                                    hgvs_p = allele["hgvs"][0]
+                                    break
+
+                # TODO should we check that assay level hgvs mtches either g. or p.?
```
bencap: Could be a good check if we want to fail the population if they don't match. If we are fine populating when they don't match, we can skip it.

bencap (commented):
Thanks for working on the scripts, and let me know if you have any questions about refactoring to share these methods with a worker process.

PR #625, opened 2025-12-18
Make the updating keywords in the editing experiment accept insensible searching

EstelleDa:
And add some controlled keywords' tests.


bencap (approved):
Thanks Estelle, looks great!

PR #628, opened 2025-12-22
Modify the Gene Ontology related validators and tests

EstelleDa:
Only finished the back-end code. Haven't done the front-end part yet.


EstelleDa:
We temporarily accept multiple combined GO terms in the molecular mechanism assessed. We'll change the related codes when we only allow one GO term in the molecular mechanism assessed's value.

PR #661, opened 2026-02-17
fix: use na_rep instead of string 'None' for missing score and count data in CSV export

### LINKED ISSUE #636: NA Representation not Applying to Score/Count CSVs
When querying the Score and Count CSV endpoints, the NA representation is not being properly applied. This results in null values being displayed with `None` (the stringified Python type) rather than our standard null representation of NA.

To reproduce:
- Query the CSV endpoint for a score set which has null scores
- Observe these scores are represented with the string `None`

We should fix the representation and add additional tests to ensure the NA rep is applying correctly for all CSV exports.

bencap:
The scores and counts loops in variant_to_csv_row were calling str() on None values without a null check, producing the literal string "None" instead of the na_rep value ("NA"). Other column handlers (core, mavedb, vep, gnomad, clingen) already had null checks — this adds the same is_null guard to scores and counts.

Additionally, bumps Pandas to >=2.2 to resolve issues with removed setup_tools libs.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -47,15 +47,15 @@ nvim = ["neovim", "python-language-server"]
 
 [[package]]
 name = "annotated-doc"
-version = "0.0.3"
+version = "0.0.4"
 description = "Document parameters, class attributes, return types, and variables inline, with Annotated."
 optional = true
 python-versions = ">=3.8"
 groups = ["main"]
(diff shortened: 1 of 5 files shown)


jstone-dev (approved):
Looks good, and I tested with score sets like urn:mavedb:00000003-b-1 that have null scores.

PR #702, opened 2026-04-10
Add controlled keywords to score set search filter options.

### LINKED ISSUE #693: Add Controlled Keywords to Score Set Search Filter Options
### Background

Controlled keywords are attached at the **experiment level** (via `experiment_controlled_keywords` bridge table), not directly on score sets. However, score sets are already searchable by keyword label through the `keywords` field in `ScoreSetsSearch`, which traverses `ScoreSet → Experiment → keyword_objs → ControlledKeyword`.

The filtering itself works, but the `/score-sets/search/filter-options` endpoint does not yet return available keyword options, so clients have no way to discover which keyword labels are valid filter values.

### What Needs to Be Done

**1. Implement keyword options query in `src/mavedb/lib/score_sets.py`**

In `fetch_score_sets_search_filter_options()`, query the distinct set of `ControlledKeyword` objects reachable from published score sets by joining through `ScoreSet → Experiment → ExperimentControlledKeywordAssociation → ControlledKeyword`, filtered to only published score sets to match search scope.

**2. Extend `ScoreSetsSearchFilterOptionsResponse` in `src/mavedb/view_models/search.py`**

There are three design options for how to surface keywords in the response. This is the key design decision for this issue.

- **Labels only** (`keywords: list[ScoreSetsSearchFilterOption]`) — consistent with how other filter options return flat string values and requires no new view model. The tradeoff is that `label` alone strips all semantic context (`key`, `system`, `code`, `description`), so clients cannot group, sort, or display keywords meaningfully without hardcoding that knowledge themselves.

- **Slim keyword objects** (`keywords: list[SlimControlledKeyword]`) — a new view model exposing only the fields relevant to search consumers (e.g. `key`, `label`, `system`, `description`), omitting internal fields like `id`, `special`, and dates. Gives clients enough context to group and render keywords without over-exposing the model. The tradeoff is that it still returns a flat heterogeneous list, so clients must group by `key` themselves if they want to present keywords by category.

- **Pre-separated by semantic group** — add one field per distinct `key` value (e.g. `keyword_biological_processes`, `keyword_molecular_functions`), each holding `list[ScoreSetsSearchFilterOption]`. This mirrors the existing pattern in `ScoreSetsSearchFilterOptionsResponse` where logically distinct categories have their own top-level fields, keeping the response self-describing and eliminating any client-side grouping logic. The tradeoff is that the response shape becomes data-driven: it is tied to the set of `key` values in the database, so adding a new keyword category requires a coordinated API and client change rather than being picked up automatically.

### Notes

- The search filter (`keywords` in `ScoreSetsSearch`) already exists and works — this issue is specifically about surfacing available keyword options through the filter options endpoint and adding these options to the UI.
- Keyword data lives at the experiment level by design; the traversal pattern using `ScoreSet.experiment.has(Experiment.keyword_objs.any(...))` is already established and should be reused.
- Eager loading for keywords is already set up in `search_score_sets()` via `selectinload` on `Experiment.keyword_objs` joined to `ExperimentControlledKeywordAssociation.controlled_keyword` — the filter options query should follow the same pattern.

EstelleDa:


bencap (approved):
Looks great, thanks Estelle.

src/mavedb/lib/score_sets.py
```diff
@@ -216,16 +216,18 @@ def build_search_score_sets_query_filter(
...
+                ScoreSet.experiment.has(
+                    Experiment.keyword_objs.any(
+                        ExperimentControlledKeywordAssociation.controlled_keyword.has(
+                            ControlledKeyword.label == label
+                        )
                     )
                 )
             )
-        )
+
```
bencap: Hmm, actually one thing that I noticed based on the front end. If a user searches for 'Other' (or any value which is ambiguous between keyword categories), all results will show up regardless of the intended class being searched. We may need to reqork the way that the controlled keyword search works to accept both a `value` and a `key`, similar to how you return both a `value` and a `key`. This would let us more precisely match the users' search request.

EstelleDa: Thanks for finding this! It's a serious problem.

src/mavedb/view_models/search.py
```diff
@@ -25,7 +30,7 @@ class ScoreSetsSearch(BaseModel):
     databases: Optional[list[str]] = None
     journals: Optional[list[str]] = None
     publication_identifiers: Optional[list[str]] = None
-    keywords: Optional[list[str]] = None
+    controlled_keywords: Optional[list[ControlledKeywordSearch]] = None
```
bencap: Thanks for the fix and the new shape. My one suggestion is that API consumers who don't know about this new change are going to be confused with what we did if they try to search by the old `keywords`. 

What I would suggest is adding a model validator and an issue to the backlog to remove it in 6 months or something:

```python
    # TODO#XXX - Remove validator after consumers have had a chance to update
    @model_validator(mode="before")
    @classmethod
    def reject_deprecated_keywords(cls, data):
        if isinstance(data, dict) and ("keywords" in data or "Keywords" in data):
            raise ValueError(
                "'keywords' is no longer supported. Use 'controlled_keywords' with "
                "a list of {key, label} objects to filter by specific keyword groups."
            )
        return data
```

Now if a user tries to make a request and use the old keywords parameter, they get a really clear error about why it doesn't work and what they need to do to fix it, rather than just a generic error that keywords isn't a valid property.

bencap (approved):

bencap:
This wasn't explicitly part of the issue, but it seems like while we are working on this if we should also make the same change to the experiment search, so that using the two models is a consistent experience.

EstelleDa:
> This wasn't explicitly part of the issue, but it seems like while we are working on this if we should also make the same change to the experiment search, so that using the two models is a consistent experience.

I think the keywords is more suitable so I renamed the controlled keywords to keywords. And we don't need the reject_deprecated_keywords model_validator anymore. Experiments used keywords already, not the controlled keywords.

src/mavedb/view_models/search.py
```diff
@@ -56,6 +70,7 @@ class ScoreSetsSearchFilterOptionsResponse(BaseModel):
     publication_author_names: list[ScoreSetsSearchFilterOption]
     publication_db_names: list[ScoreSetsSearchFilterOption]
     publication_journals: list[ScoreSetsSearchFilterOption]
+    controlled_keywords: list[ControlledKeywordFilterOption]
```
bencap: In that case, I think these should also be `keywords`. I like the decision for it to just be keywords-- I don't think there's any reason our internal verbiage needs to leak into the consumer models, but we should make sure we are consistent.

EstelleDa: Thanks for pointing this out!

bencap (approved):

PR #738, opened 2026-05-15
Bugfix/estelle/706/unique constraint to replaces

### LINKED ISSUE #706: Enforce unique supersession: `replaces_id` should have a unique constraint
### Problem

The `scoresets.replaces_id` column (FK to `scoresets.id`) has no uniqueness constraint, meaning multiple score sets can claim to supersede the same original. This creates a branching supersession graph where the code assumes a linear chain.

**Consequences:**

1. **Non-deterministic chain walking.** `superseding_score_set` is defined with `uselist=False`, so when multiple rows share the same `replaces_id`, SQLAlchemy silently picks one. Functions like `find_superseded_score_set_tail` and `find_publish_or_private_superseded_score_set_tail` may follow the "wrong" branch — e.g., an unpublished superseder instead of a published one.

2. **Row multiplication in queries.** Any LEFT JOIN on `superseding_score_set` produces N rows per original when N superseders exist. This was the root cause of the search pagination bug fixed in #675 where `LIMIT` was applied to multiplied rows rather than unique score sets.

3. **Inconsistent search results.** `fetch_superseding_score_set_in_search_result` walks the chain to find the "tail" (latest version) to display. With branching chains, different requests may resolve to different tail score sets depending on which superseder SQLAlchemy loaded.

### Diagnostic query

```sql
-- Find score sets superseded by more than one other score set
SELECT
    original.id AS original_id,
    original.urn AS original_urn,
    original.title AS original_title,
    COUNT(superseder.id) AS num_superseders,
    STRING_AGG(
        superseder.urn || ' (published: ' || COALESCE(superseder.published_date::text, 'no') || ')',
        ', ' ORDER BY superseder.id
    ) AS superseder_details
FROM scoresets original
JOIN scoresets superseder ON superseder.replaces_id = original.id
GROUP BY original.id, original.urn, original.title
HAVING COUNT(superseder.id) > 1
ORDER BY COUNT(superseder.id) DESC;
```

### Proposed fix

1. **Add a unique constraint on `replaces_id`.** Each score set may only be superseded by a single other score set. This makes the supersession graph a strict linear chain (A → B → C), matching what the code already assumes.

2. **Enforce head-node supersession at the API level.** When creating a new version, require that the score set being superseded is the current head of its chain (i.e., has no existing superseder). This prevents forking mid-chain and keeps the replacement semantics simple: you always supersede the latest version.

3. **Migration to clean up existing data.** Run the diagnostic query on staging/prod to identify any branching chains. Resolve them before applying the unique constraint — likely by picking the correct superseder for each original and updating or nullifying the others.

### Affected code

- Model: `src/mavedb/models/score_set.py` — `superseded_score_set_id` / `superseding_score_set` relationship
- Chain walking: `src/mavedb/lib/score_sets.py` — `find_superseded_score_set_tail`, `find_publish_or_private_superseded_score_set_tail`
- Search filter: `src/mavedb/lib/score_sets.py` — `build_search_score_sets_query_filter` (the `NOT EXISTS` fix works around the issue but doesn't prevent it)

EstelleDa:


src/mavedb/routers/score_sets.py
```diff
@@ -1613,6 +1614,17 @@ async def create_score_set(
...
+        if superseded_score_set is None or superseded_score_set.private:
+            logger.info(
+                msg="Failed to create score set; The newest version of the requested superseded score set is not accessible.",
+                extra=logging_context(),
+            )
+            raise HTTPException(
+                status_code=404,
+                detail="The newest version of the requested superseded score set is not accessible. "
+                       "It may be private and multiple score sets supersede the same score set.",
+            )
```
bencap, line 1626: I'm somewhat resistant here to do tail resolution on the score set that a user specifically requested as being the one they want to supersede. I think that instead, if a user requests to supersede a score set that already has been superseded, we should return a 409 Conflict with a clear message they cannot do that. Our logic simplifies to:

If the user requests to supersede a score set:
1. Check if the superseded score set exists and is accessible. If not, return a 404.
2. Check if the superseded score set already has a superseding score set. If yes, return a 409.
3. If we pass the above checks, we can supersede the score set and do so.

src/mavedb/routers/score_sets.py
bencap: Here, I also think we should add a try/catch for a potential race condition with the unique constraint. If a user sent off two almost concurrent requests for superseding the same score set, they might race past our checks above and fail here. That would look something like:

```
try:
     db.add(item)
     db.commit()
 except IntegrityError as e:
     db.rollback()
     if is_replaces_id_unique_violation(e):
         raise HTTPException(
             status_code=409,
             detail="The requested score set has already been superseded.",
         )
     raise
```

We'd need a new helper function to check if the integrity error is the one we expect.

bencap (changes requested):
Thanks for the work on this Estelle. A couple comments and a broad point about deployment. We'll need a deployment rule here where we clean production of these duplicate entries. I think we have one case right now where the duplicates are clearly junk score sets, and we can just delete those and retain the published successor that is meant to be the correct superseder. No action on this PR, but I want to make sure we persist that rule here so that we know to do it on deployment.

The other two comments are about clarity in our responses and error messages. I think we don't want to assume the user wants to supersede the head-- we should do what they say and raise a clear error if we cannot supersede the score set because of the constraint. Ultimately the root of this is that score set creation isn't necessarily atomic, but I think that is something we should handle separately in #759. Here, I think as long as we return a clear error to the user we can eliminate the confusing and unintentional behavior of having multiple superseders for the same score set.

src/mavedb/routers/score_sets.py
EstelleDa: I think your suggestion is better. I only considered that the users supersede a wrong one, but didn't consider some potential problems.

EstelleDa:
I found that only one score set has multiple private superseding score sets. I think we can delete the old superseding score set manually.

src/mavedb/routers/score_sets.py
```diff
@@ -1604,7 +1605,7 @@ async def create_score_set(
             db, item_create.superseded_score_set_urn, user_data, user_data, True
         )
 
-        if superseded_score_set is None:
+        if superseded_score_set is None or superseded_score_set.private:
```
bencap: Fetch score set by URN with our user data passed in is already checking this privacy aspect. So adding this additional private check will cause a failure if a user attempts to supersede a private score set they own.

bencap (approved):
Looks good, thanks Estelle!

Issue #549, opened 2025-10-23
Allow users to edit metadata for published calibrations


bencap:
When developing the new calibration model, it was expedient to disallow edits of certain non-sensitive fields for published calibrations. However, it may be the case that users wish to edit certain non-sensitive fields, such as sources or a baseline score description. We should allow them to do so.

We should:
- Allow editing of the following calibration fields:
    - `notes`
    - `baselineScoreDescription`
    - `researchOnly`
    - `thresholdSources`
    - `classificationSources`
    - `methodSources`
- Allow editing of the following functional range fields:
    - `description`

Issue #372, opened 2024-12-19
Make URN a non-Nullable field


bencap:
(at least) the score set URN is currently a nullable field, even though it is de-facto required on all DB records. This causes a minor annoyance while type checking, since mypy presumes this field has the potential to by Nonetype. 

As part of this fix, we should make sure other URN fields are non-nullable and that any associated type hints are corrected to eliminate the `Optional` keyword.

PR #769, opened 2026-06-15
Add superseded_calibration in Calibration. Modified related functions and tests.

### LINKED ISSUE #721: Allow Deprecation of Calibrations via Superseding
## Summary

Calibrations currently have no mechanism for deprecation. Score sets support deprecation by creating a new score set that supersedes the old one — setting `superseded_score_set_urn` on creation, which links the two via a `replaces_id` FK. This issue tracks adding equivalent superseding behavior to calibrations, including UI support for navigating the deprecation chain.

## Problem

When a calibration needs to be updated or replaced, there is no way to mark the old calibration as deprecated or link it to its replacement. Researchers viewing a score set may encounter stale calibrations with no indication that a newer version exists. This is inconsistent with the score set lifecycle model, which explicitly supports superseding. Additionally, there is no way for a user to discover the history of a calibration — whether it was replaced, or what it replaced.

## Proposed Behavior

A calibration may be superseded by another calibration belonging to the same score set. When a new calibration is created, an optional `supersededCalibrationUrn` field in the request body links it to the calibration it replaces. Once the superseding calibration is published:

- The superseded calibration is hidden from the default calibration list returned on the score set.
- The superseded calibration remains accessible directly by URN.
- Both `supersededCalibration` and `supersedingCalibration` are included in the full calibration response so consumers can follow the deprecation chain.
- The UI surfaces the deprecation chain, allowing users to navigate from a superseded calibration to its replacement and vice versa.

If the superseding calibration has not yet been published, both calibrations remain visible — matching the existing behavior for score sets.

## Acceptance Criteria

- [x] `ScoreCalibration` model has a nullable `superseded_calibration_id` FK column (`replaces_id`) pointing to another `ScoreCalibration`, with a corresponding `superseded_calibration` relationship and `superseding_calibration` back-reference.
- [x] A new Alembic migration adds the `replaces_id` column to the calibrations table.
- [x] `ScoreCalibrationCreate` (or equivalent input view model) accepts an optional `superseded_calibration_urn` string field, validated against `MAVEDB_CALIBRATION_URN_RE`.
- [x] Validation rejects attempts to supersede a private (unpublished) calibration with an appropriate error message, matching the behavior for score sets.
- [ ] The user must have `READ` permission on the superseded calibration to reference it.
- [x] `SavedScoreCalibration` (or equivalent response view model) includes `superseded_calibration: Optional[ShorterScoreCalibration]` and `superseding_calibration: Optional[ShorterScoreCalibration]`.
- [x] The calibration list returned with a score set filters out calibrations that have been superseded by a published calibration, matching the filter logic used for score sets.
- [ ] If the requesting user lacks `READ` permission on the superseding calibration, `superseding_calibration` is masked to `None` in the response.
- [ ] The UI displays a deprecation notice on a superseded calibration and provides a link or navigation element to the superseding calibration.
- [ ] The UI displays a "supersedes" notice on a superseding calibration and provides a link or navigation element back to the superseded calibration.
- [ ] Users can follow the full deprecation chain in the UI (e.g., A → B → C) without needing to manually look up URNs.
- [ ] Tests cover: creating a superseding calibration, verifying the superseded calibration is hidden once superseding is published, verifying it is still visible when superseding is unpublished, and verifying permission masking.

## Implementation Notes

- The deprecation model for score sets lives on `ScoreSet` as `superseded_score_set_id` (`replaces_id` in the DB) with ORM relationships `superseded_score_set` and `superseding_score_set`. Mirror this pattern on `ScoreCalibration`.
- Superseding validation for score sets is done in the `ScoreSetCreate` view model via a `@field_validator` on `superseded_score_set_urn`. Apply the same approach in the calibration input view model.
- Score set superseding validation enforces that the target must be a published URN (not a `tmp:` URN). Apply the same rule: only a published calibration URN may be superseded.
- The visibility filter for score sets uses an `or_` on `superseding_score_set.id.is_(None)` / `superseding_score_set.published_date.is_(None)`. Apply equivalent join-and-filter logic when building calibration queries.
- Calibrations do not have their own router today — they are managed through the score sets router. The superseding logic should be wired in at the point where calibrations are created or updated within that router.
- A calibration URN regex (`MAVEDB_CALIBRATION_URN_RE`) may need to be added to `src/mavedb/lib/validation/urn_re.py` if it does not already exist.
- The `ShorterScoreCalibration` response type (analogous to `ShorterScoreSet`) needs to be defined if it does not exist, containing at minimum the URN and title, so the UI can render links without fetching full calibration objects.
- The UI deprecation chain navigation should match how score set superseding is rendered — coordinate with the frontend on the expected shape of `supersededCalibration` and `supersedingCalibration` in the API response.

EstelleDa:


src/mavedb/lib/score_calibrations.py
```diff
@@ -321,13 +409,40 @@ async def create_score_calibration(
...
+def find_superseded_score_calibration_tail(
+    score_calibration: ScoreCalibration, action: Optional["Action"] = None, user_data: Optional["UserData"] = None
+) -> Optional[ScoreCalibration]:
+    while score_calibration.superseding_calibration is not None:
+        next_score_calibration_in_chain = score_calibration.superseding_calibration
+
+        # If we were given a permission to check and the next score calibration in the chain does not have that permission,
+        # pretend like we have reached the end of the chain. Otherwise, continue to the next score calibration.
+        if action is not None and not has_permission(user_data, next_score_calibration_in_chain, action).permitted:
+            return score_calibration
```
bencap, line 428: Here, I think we should `break` rather than returning the score calibration directly. Tracing the flow:

- A caller passes calibration A (private, superseded by B)
- We read A.superseding_calibration → B, and check permissions on B. The user has none
- We return A directly, with no permission check on A itself

This is not reachable today on production data because it's required a calibration or score set is published and public before it is superseded. However, if this ever changed this would silently widen permissions for superseded score calibrations without us realizing. Adding a `break` here would just route us through the bottom loop and check permissions as we expect.

src/mavedb/lib/score_calibrations.py
```diff
@@ -17,14 +21,91 @@
...
+    # published superseding version). Uses NOT EXISTS instead of LEFT OUTER JOIN to
+    # avoid row multiplication when multiple superseding versions point to the same
+    # original via replaces_id (which has no uniqueness constraint). A LEFT JOIN would
+    # produce N rows per original, all counted against the LIMIT, causing paginated
+    # searches to return fewer unique score calibrations than requested.
+    query = query.filter(~ScoreCalibration.superseding_calibration.has(ScoreCalibration.private.is_(False)))
+
+    if owner_or_contributor is not None:
+        query = query.filter(
+            or_(
```
bencap, line 53: The `or_` here admits calibrations the requesting user has no read permission on. Tracing it:

- Score set `S` has user `A` in `contributors`
- User `B`, who has no role on `S`, submits a calibration to `S`. `create_score_calibration_in_score_set` sets `investigator_provided = False`, since `B` is not a contributor, creator, or modifier
- `B` leaves it private
- `A` calls `/score-calibrations/me/search`. The second disjunct matches on contributorship, `search.private` is unset so nothing filters private rows, and `B`'s private calibration comes back

`_handle_read_action` would deny that row on `GET /score-calibrations/{urn}`: it is private, `A` is not the owner, `investigator_provided` is false so the contributor branch does not apply, and `A` is not an admin. `search_score_calibrations` never calls `has_permission`, so nothing catches it.

I think this is inherited from how the score-set search is put together rather than a mistake made here. `search_score_sets` has the same single `owner_or_contributor` parameter, but it has two callers and each one supplies the precondition that makes its own use safe: `/score-sets/search` forces `search.published = True` and rejects `published is False` outright, and `/me/score-sets/search` is safe because for score sets contributor implies read unconditionally. 

Neither condition applies to the calibration path:
- no `private` forcing on the calibration route, so private rows are returned by default
- no `investigator_provided` term. Score sets have no equivalent concept, so there was nothing in the template to copy

Rather than add the missing precondition at this call site, I would put the check inside the helper. `fetch_score_set_search_filter_options` already takes `requester` alongside `owner_or_contributor` and post-filters with `has_permission(requester, ...)`. `search_score_calibrations` collapsed those two into one parameter, so it cannot check permissions at all since it does not know who is asking. Restoring `requester` and filtering there makes it safe for the second caller too.

Either way, if we go the `Action` route, it is probably the case we don't actually need the endpoint this helper drives.

src/mavedb/lib/score_calibrations.py
```diff
@@ -513,13 +678,17 @@ def publish_score_calibration(db: Session, calibration: ScoreCalibration, user:
...
-    if not calibration.private:
+    if calibration.private is False: # It's possible None value
         raise ValueError("Calibration is already published.")
 
     calibration.private = False
     calibration.modified_by = user
 
     db.add(calibration)
+
+    if calibration.superseded_calibration and calibration.superseded_calibration.primary:
```
bencap, line 689: This will raise when the superseding calibration is research-use-only. `promote_score_calibration_to_primary` rejects RUO calibrations, so publishing an RUO calibration that supersedes the current primary fails here. After line 685 has already set `private = False` and line 687 has added it to the session, the request aborts partway through the change. The error the caller gets is about promotion rather than publication as well.

Publishing an RUO calibration that supersedes a primary one seems like something we should allow, so I think maybe we want to skip the auto-promote for RUO calibrations.

src/mavedb/lib/score_calibrations.py
```diff
@@ -652,6 +825,35 @@ def delete_score_calibration(db: Session, calibration: ScoreCalibration) -> None
...
+    superseded_calibration = (
+        db.query(ScoreCalibration)
+        .where(ScoreCalibration.urn == calibration_create.superseded_calibration_urn)
+        .one_or_none()
+    )
+
+    if superseded_calibration is None:
+        raise ValueError("Superseded calibration does not exist.")
+
+    if not has_permission(user_data, superseded_calibration, Action.READ).permitted:
```
bencap, line 845: This will gate on `Action.SUPERSEDE` rather than `Action.READ`.

As written, READ passes for any published calibration, so any authenticated user can supersede any published calibration in MaveDB.

So a stranger can take a calibration off a score set and strip its primary status in one request.

We already state the opposite policy one function over. `_handle_change_rank_action` carries "Community calibration owners may not ... the score set team controls ranking of community contributions." Reordering someone else's calibration is gated to the score set team, but as written deprecating it outright is gated only on being able to see it.

`Action.UPDATE` is not the substitute, which I think is why READ ended up here. `_handle_update_action` denies updates on published calibrations to everyone except admins, and only published calibrations can be superseded, so requiring UPDATE would make supersession admin-only. That is the argument for a separate action rather than reusing an existing one.

For comparison, the score set path gates supersession on owner-or-contributor rather than READ — `fetch_score_set_by_urn(db, urn, user_data, user_data, True)`, where the fourth argument requires the caller to own or contribute to the result. Worth routing score sets through the new action too so there is one spelling of this, but that can be a follow-up issue rather than part of this PR, see #854.

src/mavedb/lib/score_calibrations.py
```diff
@@ -652,6 +825,35 @@ def delete_score_calibration(db: Session, calibration: ScoreCalibration) -> None
...
+        raise ValueError("No access right to supersede this calibration.")
+
+    if superseded_calibration.private:
+        raise ValueError("Cannot supersede a private calibration. Please edit it instead.")
+
+    if superseded_calibration.superseding_calibration:
+        raise ValueError("Cannot supersede a superseded calibration. Please edit it instead.")
+
+    return superseded_calibration
```
bencap, line 854: There is no constraint that the superseded calibration belongs to the same score set as the one being created. The lookup above is by URN across the whole table, and nothing compares `superseded_calibration.score_set_id` against `containing_score_set.id`.

#721 describes supersession as happening within a score set, and the rest of the code assumes that. Two places would go wrong if a calibration on score set X supersedes one on score set Y:

- the visibility filter in `get_score_calibrations_for_score_set` is scoped by `score_set_id`, so the superseded calibration disappears from Y's list because of a row that lives on X. Nothing visible from Y explains where it went.
- `promote_score_calibration_to_primary` demotes within `score_set_id == calibration.score_set_id`, so publishing the superseding calibration on X demotes X's primary. The superseded one stays `primary = True` on Y while being filtered out of Y's list, which leaves Y with a hidden primary and no visible one.

I think this wants an explicit check, rejecting when `superseded_calibration.score_set_id != containing_score_set.id`. That does mean the function needs the containing score set passed in: right now it only sees `calibration_create` and the score set is resolved by the caller.

There is a genuine case for crossing score sets: porting calibrations onto a superseding score set, where when a score set supersedes another score set and each new calibration would then supersedes its original. That seems complicate though, so it's a separate issue: #855.

src/mavedb/models/score_calibration.py
```diff
@@ -60,6 +60,17 @@ class ScoreCalibration(Base):
...
     calibration_metadata = Column(JSONB(none_as_null=True), nullable=True)
 
+    superseded_calibration_id = Column("replaces_id", Integer, ForeignKey("score_calibrations.id"), index=True, nullable=True)
+    superseded_calibration: Mapped[Optional["ScoreCalibration"]] = relationship(
+        "ScoreCalibration",
+        uselist=False,
+        foreign_keys="ScoreCalibration.superseded_calibration_id",
+        remote_side=[id],
+    )
+    superseding_calibration: Mapped[Optional["ScoreCalibration"]] = relationship(
```
bencap, line 70: `superseding_calibration` is declared `uselist=False`, but `replaces_id` only gets a non-unique index in the migration, so nothing stops two calibrations pointing at the same original. When that happens SQLAlchemy emits `Multiple rows returned with uselist=False` and picks one of them arbitrarily, which makes the visibility filter in `get_score_calibrations_for_score_set` nondeterministic: the superseded calibration appears or disappears between requests depending on which superseding row got loaded.

We should use the behavior you implemented for score sets in #706.

src/mavedb/view_models/score_calibration.py
```diff
@@ -477,11 +479,37 @@ def functional_classifications_require_publication_sources(
...
     evidence_sources: Sequence[PublicationIdentifierCreate]
     method_sources: Sequence[PublicationIdentifierCreate]
 
+    @field_validator("superseded_calibration_urn")
+    def validate_superseded_calibration_urn(cls, v: Optional[str]) -> Optional[str]:
+        if not v:
+            return None
+
+        if urn_re.MAVEDB_CALIBRATION_URN_RE.fullmatch(v) is None:
+            if urn_re.MAVEDB_TMP_CALIBRATION_URN_RE.fullmatch(v) is None:
```
bencap, line 494: I don't think this branch is reachable. If it's written because of MyPy complaints, then we can just type ignore the `None` check and add a comment mentioning #372.

src/mavedb/routers/score_sets.py
```diff
@@ -604,7 +604,18 @@ async def fetch_score_set_by_urn(
...
-    item.score_calibrations = [sc for sc in item.score_calibrations if has_permission(user, sc, Action.READ).permitted]
+    visible_calibrations = [sc for sc in item.score_calibrations if has_permission(user, sc, Action.READ).permitted]
+
+    superseded_ids = [sc.superseded_calibration_id for sc in visible_calibrations if sc.superseded_calibration_id is not None]
+
+    available_calibrations = [sc for sc in visible_calibrations if sc.id not in superseded_ids]
+
+    # Solve Pydantic model validation error
+    for sc in available_calibrations:
+        sc.superseded_calibration = None
```
bencap, line 615: These are always null in the response now, for every caller including admins. Same loop in get_score_calibrations_for_score_set, around line 151. They're the two fields #721 adds so consumers can follow the deprecation chain, so that part of the API isn't observable. That is probably why the front end half is editor-only, since there's nothing arriving for a deprecation notice to render from.

Assigning to superseding_calibration also isn't a serialization no-op, and this is an idiom that exists in the code base but I'd like to start avoiding going forward. It's the back_populates reverse side of the many-to-one, so None detaches the other calibration and marks that row's replaces_id to be nulled. Essentially if we called `.commit()` here, we'd erase the calibrations from the database on accident even though our intention is to just hide them from the API response.

What's the Pydantic error the comment refers to? If it's ShorterScoreCalibration failing to validate against the ORM object I'd rather fix it there than drop the data.

bencap (commented):
Basically what we talked about on Slack, where we add a new action for this which belongs in `lib/permissions/score_calibration.py`. Filed as #854 so the score set half can follow separately.

We should make supersession a first-class action here rather than checking `READ` at the call site. Something like

```python
Action.SUPERSEDE: _handle_supersede_action,
```

with the predicate being the UPDATE population minus the published gate: admin, or owner, or `investigator_provided` and a contributor to the score set.

The reason I would rather it live here than inside `validate_superseded_score_calibration` is that `permissions/user-is-permitted/{model_name}/{urn}/{action}` already serves `score_calibration` and takes the `Action` enum directly, and the front end already wraps it in `useDatasetPermissions`. So adding the member is the whole change as far as the UI is concerned. The calibrations page adds one string to its actions array and can render a supersede affordance. We wouldn't need a new endpoint or a new field on the model.

The creation flow is the other case: when you are creating a calibration and want to offer the option to supersede an existing one, you need a list rather than one lookup, and `useDatasetPermissions` resolves a single URN per call. I think that wants the existing per-score-set list endpoint rather than a search, see the comment on the search filter.

We can leave the #549 published-immutability rule for now. Superseding becomes the supported way to correct a published calibration, so there is no pressure to open up updates on published rows to make this feature work.

src/mavedb/lib/score_calibrations.py
EstelleDa: I agree `break` is better. The `find_superseded_score_set_tail` function has the same code. Should we also modify that part in a separate PR? I can do it after finishing this PR.

src/mavedb/lib/score_calibrations.py
EstelleDa: It's quite helpful and clear to me. I'm thinking whether we still need the `search_my_score_calibrations` API router function. When we modified the dropdown menu to get the score calibration from the score set rather than searching for my score calibration, none of any parts calls the `search_my_score_calibrations` function. In the dashboard, users can already see all their calibrations.

src/mavedb/lib/score_calibrations.py
EstelleDa: I have already modified the calibration part. Maybe another PR for the score set's.

src/mavedb/view_models/score_calibration.py
EstelleDa: I copied this part from the view model of score set, but I also think we need this part. It's optional so it's probably `None` value. If it's not `None` value, then we need to check whether it's a temporary or published URN. Users can create score calibration through API so that they may provide a temporary URN directly.

src/mavedb/routers/score_sets.py
EstelleDa: Maybe modifying some other parts, this one is also fixed. I can't reproduce the error so I'll delete this part.

PR #774, opened 2026-06-18
feat: Migrate CAR/LDH jobs onto Allele data model

### LINKED ISSUE #741: feat: Pre-register coding and genomic Allele rows with ClinGen
## Context

Depends on: [#740](https://github.com/VariantEffect/mavedb-api/issues/740)

Coding and genomic `Allele` rows (produced by the updated mapper) must be registered with the ClinGen Allele Registry before any translation lookup runs. Because `Allele` rows are shared across score sets by VRS digest, a given allele may already have a `clingen_allele_id` from a previous score set's registration — this must be treated as a no-op.

## Goal

Add a worker job that submits unregistered `Allele` rows where `level = 'coding'` or `level = 'genomic'` to the ClinGen Allele Registry, stores the returned CA ID on the allele row, and skips rows that already have a `clingen_allele_id` set.

## Acceptance Criteria

- Worker job submits HGVS strings for alleles (`level = 'coding'` or `level = 'genomic'`) that lack a `clingen_allele_id`
- Returned CA IDs are stored on the respective `Allele` row
- Alleles that already have a `clingen_allele_id` (registered via a prior score set) are skipped without error
- ClinGen conflict responses (allele already present in registry) are handled gracefully — CA ID is extracted from the response and stored
- ClinGen API failures are retried with backoff; failures recorded via `AnnotationStatusManager` on the `allele_id`
- Score sets with no unregistered alleles complete without error

bencap:


bencap:
favor #791

Issue #775, opened 2026-06-22
chore: Backfill new annotation tables, cut over reads, and drop frozen old tables


bencap:
## Context

Depends on: #742 (annotation pipeline migrated to the Allele model), #747 (existing `MappedVariant` data backfilled to `MappingRecord` + `Allele`)

Epic #742 moved the annotation jobs onto the new Allele-based parallel-tables model behind an explicit **frozen old tables** invariant: new score sets write only to the new tables, while the old tables are read by serving for existing data and never written for new data.

That invariant was always meant to be temporary. Once the new tables exist (all 5 steps of #742) **and** existing data has been migrated into them (#747), the old tables are dead weight: they confuse the data model, double the surface a reader has to reason about, and keep the serving layer reading from a representation we no longer write. This issue is the deliberate teardown.

> Note: #747's Phase 1 describes "repoint the `gnomad_variants`/`clinical_controls` M2M associations to `allele_id`." The design diverged after #747 was written — the new model uses **dedicated ValidTime link tables** (`gnomad_allele_links`, `clinvar_allele_links`, `vep_allele_consequences`), not repointed M2M associations. The backfill must populate those link tables; this cleanup then drops the old associations. Worth reconciling #747's wording when it's picked up.

## Goal

After backfill, retire the frozen old annotation tables/columns and cut the read path over to the new tables, so the new Allele-based representation is the single source of truth.

### 1. Backfill the new annotation link tables from old data
One-time, re-runnable migration populating the new tables from existing `MappedVariant`-era data (depends on #747 having created the `Allele`/`MappingRecord` rows to link to):
- `gnomad_allele_links` ← `gnomad_variants_mapped_variants`
- `clinvar_allele_links` ← `mapped_variants_clinical_controls`
- `vep_allele_consequences` ← `MappedVariant.vep_functional_consequence` / `vep_access_date`
- Verify `Allele.hgvs_g/c/p` and `Allele.clingen_allele_id` are populated for migrated alleles (these replace the retired HGVS and variant-translation jobs).

### 2. Read-cutover
- Update `v_variant_annotations` (`api/src/mavedb/models/variant_annotation_view.py`) to project from `Allele` + the new link tables instead of `MappedVariant.hgvs_g/c/p` and the old annotation columns/associations.
- Audit any other serving queries / endpoints that read the old annotation tables and repoint them.

### 3. Drop the old tables/columns (after backfill + cutover verified in prod)
[...]

Issue #765, opened 2026-06-07
feat: skip MappingRecord re-mint when the mapping is unchanged (idempotent re-map)


bencap:
The mapping job retires and recreates the `MappingRecord` (and cascade-retires its links) on **every** re-map, even when the result is byte-identical. This is deferred deliberately — but worth tracking, because the cost is now understood to be more than storage.

## Two costs of always re-minting

1. **Storage churn** — a new `mapping_records` row (+ links) per re-map regardless of change. Benign, retroactively collapsible.
2. **Cross-level annotation availability** — re-minting cascade-retires the prior record's derived links, and the new record's derivations don't exist until the *separate* reverse-translation job reruns. So a re-map blanks a variant's cross-level derivations for the pipeline window (≈ minutes when the jobs run back-to-back). This is *correct* when the mapping actually changed (the old derivations are stale and must be recomputed) and *pure loss* when it didn't. The authoritative link is unaffected (the mapping job recreates it in the same iteration); only the reverse-translation-produced derived links blink out.

## Proposed fix

When a re-map produces the same **post-mapped allele digest** (+ `alignment_level`, `vrs_version`), keep the existing record instead of superseding it — its derived links stay live, eliminating both costs for the no-op case. Key equality on the digest (already a content hash), **not** a hand-listed set of "semantic" columns, which silently rots when a field is added (under-specify → stale data served; that asymmetry is why this was deferred rather than done eagerly).

## Non-goals

- Changing behavior when the mapping genuinely changes — re-mint + recompute is correct there.

## Context

Epic #746 (Better Reverse Translation); relates to #740 (Persist MappingRecord and Allele rows).

Issue #764, opened 2026-06-05
Decide how to register multi-variant cis-phased blocks with ClinGen


bencap:
## Context

`get_hgvs_from_post_mapped` (`src/mavedb/lib/variants.py`) now supports an opt-in `combine_cis=True` that collapses a multi-variant cis-phased block (VRS `Haplotype` / `CisPhasedBlock`) into a single bracketed HGVS expression, e.g. `NC_000001.11:g.[123A>G;125T>C]`, via the new `join_cis_phased_hgvs` helper in `src/mavedb/lib/hgvs.py`.

We have enabled `combine_cis=True` in:
- the variant mapping job (`hgvs_assay_level` + `Allele.hgvs_*` population)
- the score-set response/export view models (`post_mapped_hgvs_g` / `post_mapped_hgvs_p` fallbacks)

We have **deliberately left it disabled** for ClinGen Allele Registry submission (`src/mavedb/worker/jobs/external_services/clingen.py`, both call sites). Today these variants return no HGVS and are recorded as `MISSING_IDENTIFIER` failures and skipped.

## The question

A multi-variant cis-phased block has no single Canonical Allele ID, and it is unclear whether the ClinGen Allele Registry accepts a bracketed cis-phased HGVS expression (`g.[a;b]`) at all, or whether these should instead be:

- registered per-component (one CAID per member), then associated, or
- represented some other way the registry prefers, or
- intentionally left unregistered.

## Action

Work with ClinGen to confirm the intended handling, then either enable `combine_cis=True` at the ClinGen call sites (and adjust submission/annotation accordingly) or formalize the skip. Update the inline notes referencing this issue once resolved.

Issue #747, opened 2026-05-18
feat: Retroactive backfill — migrate MappedVariant data and run reverse translation for existing score sets


bencap:
## Context

Depends on: [#742](https://github.com/VariantEffect/mavedb-api/issues/742)

The new mapping pipeline covers score sets going forward. Existing score sets in MaveDB have `MappedVariant` records under the old schema. A two-phase backfill is needed: (1) migrate all existing `MappedVariant` rows to the new `MappingRecord` + flat `Allele` schema, and (2) run reverse translation for protein-level score sets that currently lack coding and genomic `Allele` rows.

## Goal

A one-time (re-runnable) backfill job that migrates existing `MappedVariant` data to the new schema and ensures all protein-level score sets have a full complement of coding and genomic `Allele` rows with ClinGen registration and annotations.

## Phase 1 — Schema migration of existing data

For each existing `MappedVariant`:
- Create a `MappingRecord` row carrying provenance fields (`pre_mapped`, `mapping_api_version`, `mapped_date`, `current`, QC fields)
- Create `Allele` rows from existing `hgvs_g`, `hgvs_c`, `hgvs_p` columns, populating `level`, `transcript`, and `post_mapped` as available (upsert by VRS digest)
- Create `mapping_record_alleles` association rows
- Migrate `VariantAnnotationStatus` rows from `variant_id` to `allele_id`
- Repoint `gnomad_variants` and `clinical_controls` M2M associations from `mapped_variant_id` to `allele_id`

## Phase 2 — Reverse translation for protein-level score sets

For each protein-level score set with `MappingRecord` rows but no associated coding/genomic `Allele` rows:
- Run reverse translation to enumerate all coding variants
- Upsert resulting `Allele` rows (`level = 'coding'` and `level = 'genomic'`) and link via `mapping_record_alleles`
- Run ClinGen pre-registration (#741) and annotation (#742) for new alleles

## Open Question: `variant_translations` deprecation

Once all score sets have `Allele` rows at all levels, the `variant_translations` flat lookup table may be redundant. Whether the FK join (`AssayedVariant → MappingRecord → Allele`) is fast enough for search at scale should be benchmarked during the backfill. If performant, `variant_translations` can be deprecated; if not, it stays as a denormalized search index populated from `Allele` FK data.

## Acceptance Criteria

- Phase 1 migrates all existing `MappedVariant` rows to `MappingRecord` + `Allele` rows without data loss
[...]

PR #783, opened 2026-06-30
feat: Migrate the annotation pipeline onto the allele model + append-only event log

### LINKED ISSUE #742: feat: Extend annotation pipeline to cover Allele entities
## Context

Depends on: [#741](https://github.com/VariantEffect/mavedb-api/issues/741)

`Allele` rows at each level are real variants that can be independently annotated. Because `Allele` rows are shared across score sets by VRS digest, annotation results are stored once per allele and shared across all score sets that reference it. Annotation jobs must check for an existing current annotation before re-annotating.

## Annotation architecture

`AnnotationStatus` is QC/audit only — it tracks whether the pipeline ran and whether it succeeded. The actual annotation data lives in dedicated per-type tables (`VEPAnnotation`, `GnomADVariant`, `ClinicalControl`), each with `superseded_at` for temporal support.

When a new annotation is produced for an allele that already has a current annotation of that type, the existing row's `superseded_at` is set to `NOW()` in the same transaction that inserts the new row. This ensures the temporal query `WHERE created_at <= T AND (superseded_at IS NULL OR superseded_at > T)` always returns exactly one result for any point in time.

## Level-to-annotation routing

All routing is driven by the `level` column on the flat `alleles` table:

- **gnomAD allele frequency** → `level = 'genomic'` → stored in `GnomADVariant`
- **VEP functional consequence** → `level = 'genomic'` or `level = 'coding'` → stored in `VEPAnnotation`
- **ClinVar / clinical controls** → `level = 'coding'` or `level = 'protein'` → stored in `ClinicalControl`
- **ClinGen allele ID** → stored as `clingen_allele_id` column on the `alleles` table (stable, no temporal table needed)

## Acceptance Criteria

- VEP annotation job creates `VEPAnnotation` rows linked by `allele_id`; sets `superseded_at` on the previous current row in the same transaction
- gnomAD annotation job creates `GnomADVariant` rows linked by `allele_id`; sets `superseded_at` on the previous current row in the same transaction
- ClinVar annotation job creates `ClinicalControl` rows linked by `allele_id`; sets `superseded_at` on the previous current row
- `AnnotationStatus` rows are created per job run per `allele_id` for QC tracking only
- Annotation jobs skip alleles that already have a current annotation of the same type and source version (cross-score-set deduplication)
- Temporal query pattern is verified: querying annotation state at a past timestamp returns the correct historical annotation
- Existing `AssayedVariant`-level annotation behavior is unchanged

bencap:
### Summary

Builds on the deduplicated allele / valid-time mapping model from #741 (and #739). This PR moves the **external-annotation pipeline** onto that model and replaces per-variant annotation *status* with an append-only event log.

Two themes:
1. **Annotation jobs link to alleles, not `MappedVariant`.** gnomAD, VEP, and ClinVar now resolve and link against the deduplicated allele space (authoritative *and* RT-derived alleles), with version-keyed refresh and idempotent re-runs. CAR/LDH already ran on alleles.
2. **Status becomes an append-only `AnnotationEvent` log.** A single log across the pipeline records *disposition + why + when* per subject; "current" is derived (`v_current_annotation_events`). This can distinguish confirmed-absence from never-checked — the gap `VariantAnnotationStatus` couldn't express.

Still parallel-tables: new link tables and the event log are written for new runs; legacy `MappedVariant` columns and association tables stay frozen-serving. Backfill/read-cutover/drops are deferred to #775/#747.

### Annotation jobs onto the allele model

- **gnomAD** → `gnomad_allele_links` (one live link per allele; supersede-on-change; version-keyed skip with a `force` bypass).
- **VEP** → `vep_allele_consequences` (collapsed record+link; keyed on the Ensembl release; supersede only on a *changed* consequence so categorical values don't churn each release). HTTP rejections are routed to Variant Recoder rather than failing the batch.
- **ClinVar** → `clinical_controls` renamed to `clinvar_controls` (ORM/table/constraint only; the `ClinicalControl*` serving contract is intentionally frozen) + multi-live `clinvar_allele_links` (one live link per archival release; get-or-create). Adds forward-captured, unserved `clinvar_variation_id`.
- All consumers link **every** current allele via the shared `group_alleles_for_annotation` primitive (one work-unit per allele); `force_reregister` re-registers existing CAIDs.
- New `lib/alleles.py::get_allele_translations` — cross-layer equivalence via mapping-record co-membership, with `as_of` temporal support.

### Append-only annotation event log

- New **`annotation_event`** table + **`v_current_annotation_events`** view (latest-event-per-subject). `AnnotationStatusManager` rewritten to emit events; mapping, reverse-translation, gnomAD, VEP, ClinVar, and CAR/LDH all emit events from their resolution outcomes.
- Records **status, not value**: `disposition` (`present`/`absent`/`not_applicable`/`failed`; `pending` = no event) + a granular per-domain `reason`. Events share the job's DB session (flushed-not-committed), so they're transactionally coupled to the annotation data they describe.
- Adds a pipeline-tracking + event-aware annotations script.

### Retired

- Deleted the `populate_hgvs_for_score_set` and `populate_variant_translations_for_score_set` jobs (`hgvs.py`, `variant_translation.py`) — both DAG leaf nodes, no edge repair needed. `Allele.hgvs_*` is mapping-populated; the RT equivalence space replaces variant translations. `lib/variant_translations.py` + the `variant_translations` table stay frozen (serving-only).

### Migrations

8 new (the three annotation link/consequence tables, the `clinvar_controls` rename, `clinvar_variation_id`, the event table + view, and an index on `annotation_event.score_set_id`) plus a rename of the existing alleles/mapping-records migration to match its contents. Single alembic head; all have working downgrades.

### Testing

`mypy src/` and `ruff` green. Updated/added coverage across `tests/{lib,models,worker,routers}` for each migrated job, the event model + view, and the rewritten status manager. (API/worker tests per project convention; UI out of scope.)

### Deferred follow-ups

- **#765** — digest-keyed mapping idempotency (closes the brief re-map window where derived-allele annotations are unreachable from the variant).
- **#775 / #747** — backfill from legacy data, read-cutover, drop frozen tables; wire the new view into serving (the "current-for-variant" walk is deliberately not built yet).
- **#764** — ClinGen submission has no single CAID for a multi-variant cis block.
- `clinvar_controls`→`clinvar_variants` serving remodel (expose VariationID) — deferred to read-cutover.
- Materialize `v_current_annotation_events` + add a retention/compaction sweep **if/when** cross-subject filtering or log growth becomes hot (fine at current scale; read path is internal-only).


bencap:
favor #791

PR #785, opened 2026-07-07
feat(uniprot): support Ensembl protein accessions in UniProt ID mapping

davereinhart:
infer_db_name_from_sequence_accession now recognizes ENSP accessions and maps them to the Ensembl_Protein source, so accession-based target genes with Ensembl protein IDs resolve to a UniProt ID instead of raising NotImplementedError.

Also make validate_ensembl_identifier accept versioned Ensembl stable IDs (e.g. ENSP00000369497.3) by stripping the trailing .<version> before the idutils.is_ensembl check, matching how validate_refseq_identifier already handles versioned RefSeq accessions.

Related to varianteffect/dcd_mapping2#110

### DIFF
--- a/src/mavedb/lib/validation/identifier.py
+++ b/src/mavedb/lib/validation/identifier.py
@@ -1,3 +1,5 @@
+import re
+
 import idutils
 
 from mavedb.lib.validation.constants.identifier import valid_dbnames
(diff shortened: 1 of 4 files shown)


src/mavedb/lib/validation/identifier.py
```diff
@@ -62,7 +67,8 @@ def validate_ensembl_identifier(identifier: str):
     ValidationError
         If the identifier is not a valid Ensembl identifier.
     """
-    if not idutils.is_ensembl(identifier):
+    base_identifier = re.sub(r"\.\d+$", "", identifier)
```
davereinhart, line 70: We can remove this line after updating idutils if/when this change gets incorporated:
https://github.com/inveniosoftware/idutils/pull/149

bencap (approved):

PR #847, opened 2026-08-17
fix: refget endpoint reported an incorrect ``Content-Length``, leading to ``RuntimeError``s

### LINKED ISSUE #846: Refget sequence endpoint declares the full sequence length as `Content-Length` on subsequence requests
## Summary

`GET /api/v1/refget/sequence/{alias}` sets `Content-Length` to the length of the whole sequence on every response, including requests that ask for a subsequence. The streamed body is then shorter than the declared length, and the ASGI server aborts the response with `RuntimeError: Response content shorter than Content-Length` after the status line is already committed. Clients receive a truncated body.

## Problem

In the refget router's `get_sequence`, the response headers are built from the sequence metadata rather than from the requested interval:

```python
headers = {"Content-Length": str(seqinfo["len"])}
```

For `GET /api/v1/refget/sequence/NC_000016.10?start=2077608&end=2085800`, the body is 8,192 bytes (one `DEFAULT_CHUNK_SIZE` chunk from `sequence_generator`) while the header declares 90,338,345. Uvicorn raises when the stream ends short.

The traceback surfaces inside `CatchAllErrorMiddleware`, but the middleware is not at fault: `response_started` is already `True`, so it re-raises rather than converting a committed 200 into a 500. The failure is in the header the route set.

The defect is masked in two places:

- `GZipMiddleware` strips `Content-Length` from compressed responses, so any client sending `Accept-Encoding: gzip` never sees the mismatch. Only clients requesting identity encoding (e.g. curl without `--compressed`) hit it.
- `TestClient` sends `Accept-Encoding: gzip` by default, so no existing test could observe the header.

A second defect shares the same root: coordinate validation is gated on `start is not None and end is not None`, so a request supplying only one bound is never bounds-checked. `?start=99999999` alone streams an empty body under a 200 whose `Content-Length` is the full sequence length — the same crash by a different route.

The seqrepo router's `get_sequence` does not set `Content-Length` and therefore cannot produce this crash, but it shares the validation gap. Its only check is `start > end` → 422. Coordinates past the end of the sequence, and negative coordinates, are passed to `sequence_generator` unchecked and return a truncated or empty body under a 200.

## Steps to reproduce

1. Request a subsequence with identity encoding:

```
curl -H 'Accept-Encoding: identity' \
  'https://api.mavedb.org/api/v1/refget/sequence/NC_000016.10?start=2077608&end=2085800'
```

2. Observe the server-side `RuntimeError` and the client-side incomplete read.

**Expected:** 200 with an 8,192-byte body and `Content-Length: 8192`.

**Observed:** `RuntimeError: Response content shorter than Content-Length` on the server; the client reads 8,192 bytes and reports `IncompleteRead(8192 bytes read, 90330153 more expected)` against a declared `Content-Length: 90338345`.

Requesting the same URL without `Accept-Encoding: identity` succeeds, because gzip removes the header.

## Proposed behavior

Both sequence routes resolve the requested interval to concrete half-open coordinates once, immediately after the sequence length is known, and derive validation, headers, and the streamed body from those same values.

Refget route:

- `Content-Length` equals the number of bytes actually streamed (`end - start`), not the sequence length.
- Bounds validation runs whenever either bound is supplied, not only when both are.
- `Content-Range` and the generator bounds use the resolved coordinates.
- Out-of-range coordinates return 416 with a `Content-Range: bytes */{length}` header, matching the existing two-sided behavior.

Seqrepo route:

- Bounds validation runs whenever either bound is supplied.
- Out-of-range or negative coordinates return 422, consistent with the 422 this route already returns for `start > end`. This route has no `Range` header, `Content-Range`, or `Content-Length` semantics, so 416 would be the wrong code here even though refget uses it.

## Acceptance criteria

- For every request shape against the refget sequence route — no bounds, both bounds, start only, end only, and a `Range` header — the returned `Content-Length` equals the byte length of the returned body. Tests must send `Accept-Encoding: identity`, or `GZipMiddleware` will strip the header and the assertion will not exercise the fix.
- A `Range: bytes=1-2` request against a 4-base sequence returns 206 with `Content-Range: bytes 1-2/4` and `Content-Length: 2`.
- Refget requests supplying only `start`, or only `end`, return the correct subsequence with a 200.
- Refget requests supplying only `start` beyond the sequence length, or only `end` beyond the sequence length, return 416 with a `Content-Range` header. These currently return an empty 200.
- Seqrepo requests supplying only `start`, or only `end`, return the correct subsequence with a 200.
- Seqrepo requests with coordinates outside the sequence — `start=10&end=12`, `start=1&end=12`, `start=10`, `end=12`, `start=-1&end=2` against a 4-base sequence — return 422 with `Invalid coordinates` in the detail. These currently return a truncated or empty 200.
- Existing refget and seqrepo router tests continue to pass unchanged, including the two-sided 416 cases and the seqrepo `start > end` 422.

## Implementation notes

- The shared `sequence_generator` helper in the seqrepo lib module already normalizes `None` bounds internally. Passing it resolved coordinates instead is equivalent but makes the length contract explicit at the call site, and removes the possibility of the route and the generator disagreeing about what interval is being served.
- Refget's `start > end` check raises 501 (circular chromosomes unsupported) and runs before the alias lookup, so invalid coordinates are rejected even for unknown aliases. Keep that ordering; only the length-dependent checks need to move below the lookup.
- Once bounds are resolved, the 206 branch condition can be reduced to `if range_header:` — the `Range` header regex requires both bounds, so the extra `start is not None and end is not None` conjuncts are always true when a header is present.
- `sequence_generator` breaks out of its loop on an empty chunk. For a validated range this can only happen on corrupt or missing seqrepo data, but it would silently truncate the body and reintroduce the `Content-Length` mismatch. Out of scope here; worth tracking separately if this error recurs on a range that passes validation.
- The refget sequence route is the only place in the codebase that sets an explicit `Content-Length` on a `StreamingResponse`. The NDJSON export routes in the score sets router stream without one and are unaffected.
- The seqrepo sequence route has no callers in the UI beyond its generated OpenAPI type, so the 200 → 422 change for out-of-range coordinates has a small blast radius.

bencap:
## Summary

The refget sequence endpoint declared `Content-Length` as the full length of the sequence on every response, including subsequence requests. `NC_000016.10?start=2077608&end=2085800` streamed 8,192 bytes under a declared 90,338,345, and the ASGI server aborted the response with `RuntimeError: Response content shorter than Content-Length` after the status line was already committed. Clients got a truncated body.

Both sequence endpoints now resolve `start`/`end` to concrete half-open bounds once the sequence length is known, and derive validation, headers, and the streamed body from those same values.

## Changes

- **refget:** `Content-Length` reports the bytes actually streamed. Bounds validation runs whenever either bound is supplied, not only when both are. A one-sided out-of-range request previously skipped validation entirely and returned an empty 200 carrying the full sequence length.
- **seqrepo:** same one-sided bounds check. This endpoint never set `Content-Length` and so could not crash, but returned a truncated body under a 200 for coordinates past the end of a sequence.
- **tests:** every request shape asserts `Content-Length == len(body)`; added coverage for service-info and the range/query-param conflict; sharpened the `unit`/`integration` marker definitions.

## Behavior changes

Requests that were broken and now work:

| Request | Before | After |
| --- | --- | --- |
| refget subsequence, client not accepting gzip | truncated body, `RuntimeError` server-side | 200/206 with correct `Content-Length` |
| refget one-sided bound within range (`?start=1`, `?end=3`) | correct bytes under a full-length `Content-Length` | correct bytes and header |

Requests that were silently wrong and are now rejected:

| Request | Before | After |
| --- | --- | --- |
| refget, one-sided bound out of range | empty 200 | 416 with `Content-Range` |
| seqrepo, coordinates past end of sequence | truncated/empty 200 | 422 |
| seqrepo, negative `start` | passed through to seqrepo | 422 |

## Note

`GZipMiddleware` strips `Content-Length` from compressed responses, and `TestClient` requests gzip by default. That is why this reached production with the endpoint under test, and why the new assertions pin `Accept-Encoding: identity`. Without that header they pass regardless of the fix.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -111,8 +111,8 @@ testpaths = "tests/"
 pythonpath = "."
 norecursedirs = "tests/helpers/"
 markers = """
-    integration: tests that exercise multi-component flows and/or database interactions.
-    unit: fast unit tests with isolated dependencies.
+    integration: tests that exercise a multi-component flow end-to-end, with no internal mocking.
+    unit: fast, isolated tests. Real un-mocked collaborators are fine if they're local, fast, and deterministic.
     network: tests which interact with external services over a network connection.
     slow: tests which are slow running.
 """
--- a/src/mavedb/routers/refget.py
+++ b/src/mavedb/routers/refget.py
@@ -183,37 +183,47 @@ def get_sequence(
         )
 
     seq_id = seq_ids[0]
-    seqinfo = sr.sequences.fetch_seqinfo(seq_id)
+    seq_len = sr.sequences.fetch_seqinfo(seq_id)["len"]
 
-    if start is not None and end is not None:
-        if start >= seqinfo["len"]:
+    # Resolve to concrete half-open bounds up front so validation, Content-Length, and the streamed
+    # body all agree, even when only one of start/end is supplied.
+    seq_start = start if start is not None else 0
+    seq_end = end if end is not None else seq_len
+
+    if start is not None or end is not None:
+        if seq_start >= seq_len:
             raise HTTPException(
                 status_code=416,
                 detail="Invalid coordinates: start > sequence length",
-                headers={"Content-Range": f"bytes */{seqinfo['len']}"},
+                headers={"Content-Range": f"bytes */{seq_len}"},
             )
-        if end > seqinfo["len"]:
+        if seq_end > seq_len:
             raise HTTPException(
                 status_code=416,
                 detail="Invalid coordinates: end > sequence length",
-                headers={"Content-Range": f"bytes */{seqinfo['len']}"},
+                headers={"Content-Range": f"bytes */{seq_len}"},
             )
-        if not (0 <= start <= end <= seqinfo["len"]):
+        if not (0 <= seq_start <= seq_end <= seq_len):
             raise HTTPException(
                 status_code=416,
                 detail="Invalid coordinates: must obey 0 <= start <= end <= sequence_length",
-                headers={"Content-Range": f"bytes */{seqinfo['len']}"},
+                headers={"Content-Range": f"bytes */{seq_len}"},
             )
 
-    headers = {"Content-Length": str(seqinfo["len"])}
-    if start is not None and end is not None and range_header:
+    # Content-Length must match bytes actually streamed. Overstating it aborts the response mid-stream
+    # once the ASGI server has already committed the status line.
+    headers = {"Content-Length": str(seq_end - seq_start)}
+    if range_header:
         status = 206
-        headers["Content-Range"] = f"bytes {start}-{end - 1}/{seqinfo['len']}"
+        headers["Content-Range"] = f"bytes {seq_start}-{seq_end - 1}/{seq_len}"
         headers["Accept-Ranges"] = "bytes"
     else:
         status = 200
         headers["Accept-Ranges"] = "none"
 
     return StreamingResponse(
-        sequence_generator(sr, seq_ids[0], start, end), media_type="text/plain", status_code=status, headers=headers
+        sequence_generator(sr, seq_id, seq_start, seq_end),
+        media_type="text/plain",
+        status_code=status,
+        headers=headers,
     )
--- a/src/mavedb/routers/seqrepo.py
+++ b/src/mavedb/routers/seqrepo.py
@@ -71,7 +71,23 @@ def get_sequence(
             status_code=400, detail=f"Multiple sequences exist for alias '{alias}'. Use an explicit namespace."
         )
 
-    return StreamingResponse(sequence_generator(sr, seq_ids[0], start, end), media_type="text/plain")
+    seq_id = seq_ids[0]
+    seq_len = sr.sequences.fetch_seqinfo(seq_id)["len"]
+
+    # Resolve to concrete half-open bounds so the check and the streamed body agree, even when only
+    # one of start/end is supplied.
+    seq_start = start if start is not None else 0
+    seq_end = end if end is not None else seq_len
+
+    if start is not None or end is not None:
+        if not 0 <= seq_start < seq_len:
+            logger.error(msg="Invalid coordinates: start lies outside the sequence.", extra=logging_context())
+            raise HTTPException(status_code=422, detail=f"Invalid coordinates: must obey 0 <= start < {seq_len}")
+        if not seq_start <= seq_end <= seq_len:
+            logger.error(msg="Invalid coordinates: end lies outside the sequence.", extra=logging_context())
+            raise HTTPException(status_code=422, detail=f"Invalid coordinates: must obey start <= end <= {seq_len}")
+
+    return StreamingResponse(sequence_generator(sr, seq_id, seq_start, seq_end), media_type="text/plain")
 
 
 @router.get("/metadata/{alias}", response_model=SeqRepoMetadata, summary="Get sequence metadata by alias")
--- a/tests/routers/test_refget.py
+++ b/tests/routers/test_refget.py
@@ -3,6 +3,8 @@ from unittest.mock import patch
 
 import pytest
 
+pytestmark = pytest.mark.unit
+
 arq = pytest.importorskip("arq")
 cdot = pytest.importorskip("cdot")
 fastapi = pytest.importorskip("fastapi")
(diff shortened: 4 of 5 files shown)


jstone-dev:
Claude did a more thorough job reviewing this change than me, so I'll quote its report:

# Code Review: mavedb-api #847

**PR:** [fix: refget endpoint reported an incorrect `Content-Length`, leading to `RuntimeError`s](https://github.com/VariantEffect/mavedb-api/pull/847)
**Base:** `release-2026.3.0` ← **Head:** `bugfix/bencap/846/refget-content-length-for-subsequence-requests`
**Diff:** 5 files, +182/−15 (`refget.py`, `seqrepo.py`, 2 test files, `pyproject.toml`)
**Review method:** 4 parallel sub-agent passes (line-by-line diff scan, removed-behavior audit, cross-file trace, reuse/efficiency scan) against the checked-out PR branch

## Summary

The core fix is correct: `Content-Length` is now derived from the actual resolved byte range instead of the full sequence length, which eliminates the original `RuntimeError: Response content shorter than Content-Length` crash for subsequence requests. However, the fix introduces one new correctness bug, tightens validation in ways that silently break previously-working requests, and applies unrelated scope creep to a sibling endpoint (`seqrepo.py`) that never had the bug in the first place.

Note: the PR's own description already discloses that `GZipMiddleware` strips `Content-Length` on gzip-capable responses (confirmed empirically — Starlette always sets `more_body=True` on the first `StreamingResponse` chunk, so `minimum_size` never protects a streamed route). That means the fix only matters for identity-encoding clients (e.g. raw `curl`); this is disclosed scope, not a gap.

## Findings

### 1. Malformed Range survives every guard — `refget.py` (206 response with inverted `Content-Range`)

`Range: bytes=N-(N-1)` (e.g. `bytes=5-4`) parses to `start == end` after the header's inclusive end is converted to a half-open bound. That collapses `seq_start == seq_end`, which passes the `start > end` 501 guard *and* all three 416 checks. The endpoint returns **206** with `Content-Range: bytes 5-4/7` (last-byte-pos < first-byte-pos — invalid per RFC 9110 §14.4) and `Content-Length: 0`.

- Pre-PR: this crashed loudly (the bug being fixed).
- Post-PR: it succeeds silently with a spec-invalid header — worse for debuggability, and a caching proxy or strict client may reject or mis-cache it.
- **Root cause:** the Range-header-to-bounds conversion happens *before* the shared `start > end` check, so that check can no longer catch `start == end` from a reversed range.

### 2. Over-length ranges now rejected instead of clamped — `refget.py`

Per RFC 9110 §14.1.1, a `Range` whose last-byte-pos exceeds the resource length must be **clamped** to the actual length and served, not rejected. This PR widens the bounds guard from `start is not None and end is not None` to `start is not None or end is not None`, so a request like `Range: bytes=0-999999` (or any one-sided/over-long range) now returns **416** instead of the full/clamped content.

- No test in `test_refget.py` covers a one-sided range, so this behavior change ships uncovered either way.
- This may be an intentional tightening rather than an oversight, but it's undocumented in the PR description and is a spec deviation — worth a product decision, not a silent change.

### 3. `seqrepo.py` scope creep: new 422s on an endpoint that never had this bug

`seqrepo.py` gained the same style of bounds validation, but this router **never sets `Content-Length`** at all (still fully chunked), so it could not have produced the `RuntimeError` this PR is fixing. The new validation is an undocumented breaking change:

- The route's OpenAPI `responses` dict still declares only `200` — the new `422` isn't documented.
- Previously, `?start=<past-end>` returned `200` with an empty/clamped body; it now returns `422`. Existing clients probing past sequence end will break.

### 4. Inconsistent bounds semantics between the two sibling routers

- `seqrepo.py` uses `0 <= seq_start < seq_len` (**strict** `<`), which rejects the legitimate empty-tail slice `start == seq_len` (e.g. paginating exactly to the end of a sequence).
- `refget.py`'s equivalent check uses `>=` and returns **416** for the identical condition, while `seqrepo.py` returns **422**.
- The two endpoints now disagree on both *which coordinates are legal* and *what status code reports the same client mistake* for what should be the same underlying semantics.

### 5. Efficiency / reuse (lower priority, not a correctness bug)

- `sr.sequences.fetch_seqinfo(seq_id)` is now called **twice per request** in both routers — once in the router to get `seq_len`, and again inside `sequence_generator` (`lib/seqrepo.py`), which isn't cached across requests since a fresh `SeqRepo` instance is constructed per request.
- Start/end defaulting logic is now duplicated in three places (`refget.py`, `seqrepo.py`, `lib/seqrepo.py`'s `sequence_generator`), which retains an `Optional[int]` signature and internal defaulting that's dead code now that both call sites always pass resolved ints.
- Recommend extracting a single `resolve_bounds(sr, seq_id, start, end) -> (seq_start, seq_end, seq_len)` helper shared by both routers, and narrowing `sequence_generator` to take resolved, non-optional bounds.

## Recommendation

- **Fix before merge:** #1 (malformed-range hole) and #3/#4 (seqrepo.py scope creep and status-code/bound inconsistency with refget.py).
- **Needs a product decision, not a silent change:** #2 (clamp vs. reject for over-length ranges) — pick one behavior and add test coverage for one-sided ranges either way.
- **Optional cleanup:** #5, to remove the duplicated `fetch_seqinfo` calls and triplicated bounds logic.

PR #849, opened 2026-08-18
fix(worker): clear stale uniprot_id_from_mapped_metadata on remap

### LINKED ISSUE #794: Clear existing `uniprot_id_from_mapped_metadata` value from `target_gene` prior to updating
In cases where a previous mapping job has populated the `target_gene.uniprot_id_from_mapped_metadata` value, a subsequent mapping job may not result in a UniProtID value. In theses case the existing value can remain in this field when `target_gene` fields are updated, resulting in a mismatch that would not occur in a new `target_gene` record.

To prevent this, the `uniprot_id_from_mapped_metadata` field should be cleared when a `target_gene` is being updated as part of the `map_variants_for_score_set` function, and prior to running UniProt ID mapping.

davereinhart:
The UniProt polling job only writes target_gene.uniprot_id_from_mapped_metadata on a successful lookup. When a remap produced no UniProt ID (no results, ambiguous results, or the target gene was not found), the value from the previous mapping run survived and became mismatched against the freshly written pre/post mapped metadata — a state unreachable for a new target_gene record.

Clear the field alongside the other mapped metadata writes in map_variants_for_score_set, so it is reset before the downstream UniProt mapping jobs run.

Closes #794

### DIFF
--- a/src/mavedb/worker/jobs/variant_processing/mapping.py
+++ b/src/mavedb/worker/jobs/variant_processing/mapping.py
@@ -203,6 +203,8 @@ async def map_variants_for_score_set(ctx: dict, job_id: int, job_manager: JobMan
 
             target_gene.pre_mapped_metadata = cast(pre_mapped_metadata, JSONB)
             target_gene.post_mapped_metadata = cast(post_mapped_metadata, JSONB)
+            target_gene.uniprot_id_from_mapped_metadata = None
+
             job_manager.db.add(target_gene)
             logger.debug("Added mapping metadata to target gene.", extra=job_manager.logging_context())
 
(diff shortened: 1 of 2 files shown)


src/mavedb/worker/jobs/variant_processing/mapping.py
bencap, line 175: Looking at your changes, I am thinking we should also do a similar thing here. This is basically an identical bug to the `uniprot_id_from_mapped_metadata`. If the mapper doesn't supply `gene_info`, we retain the stale gene info from a prior run.

We could add a new issue but I was thinking it might be small and similar enough to the purpose of this bug fix that we just fold it in.

bencap (approved):
Thanks Dave, looks good. I had a thought that our handling of the `mapped_hgnc_symbol` property looked really similar to the shape of this bug. If you'd rather keep this PR clean to just the issue at hand, I can open up a new issue for that one. Just lmk.

src/mavedb/worker/jobs/variant_processing/mapping.py
davereinhart: @bencap I added a new commit to include this. Let me know if that looks ok and if you think it should have its own unit test like the cleared uniprot ID.

bencap: Another unit test would be fantastic.

davereinhart: Unit test added for clearing `mapped_hgnc_name`.

PR #859, opened 2026-08-26
build(dev): replace LocalStack with moto for local S3 emulation

### LINKED ISSUE #768: Replace LocalStack with Floci for local S3 emulation
## Summary

LocalStack's community edition was archived in March 2026, with core services (including S3) moved behind a \$39/month paywall. MaveDB uses LocalStack exclusively to emulate S3 locally for score set CSV upload and download flows. Floci is a MIT-licensed drop-in replacement that emulates AWS APIs on the same port (4566) and is compatible with boto3 and the AWS CLI.

## Problem

`docker-compose-dev.yml` currently pulls `localstack/localstack:latest`, which now requires a paid auth token to start. Developers running the full dev stack will encounter auth failures on `docker compose up` as this image no longer works without a license.

## Proposed Behavior

Replace the `localstack` service in `docker-compose-dev.yml` with an equivalent Floci service. The S3 endpoint URL, bucket name, and init script behavior should remain functionally identical from the application's perspective.

## Acceptance Criteria

- `docker compose up` starts successfully with no LocalStack auth errors
- The Floci service is reachable at `http://floci:4566` from within the compose network
- The `score-set-csv-uploads-dev` bucket is created automatically on container startup
- A score set CSV can be uploaded via the API and subsequently downloaded during variant processing without error
- `S3_ENDPOINT_URL` in `settings/.env.template` reflects the updated service hostname (`http://floci:4566`)

## Implementation Notes

**`api/docker-compose-dev.yml`**

Replace the `localstack` service block:

```yaml
localstack:
  image: localstack/localstack:latest
  ports:
    - "4566:4566"
  env_file:
    - settings/.env.dev
  environment:
    - SERVICES=s3:4566
  volumes:
    - mavedb-localstack-dev:/var/lib/localstack
    - "./bin/localstack-init.sh:/etc/localstack/init/ready.d/localstack-init.sh"
```

With a Floci equivalent (confirm the exact image name and init hook path from the Floci docs — the volume mount path for init scripts differs from LocalStack's `/etc/localstack/init/ready.d/`). Also rename the named volume from `mavedb-localstack-dev` to `mavedb-floci-dev` and update the `volumes:` block at the bottom of the file accordingly.

**`api/bin/localstack-init.sh`**

The init script currently uses the `awslocal` wrapper CLI:

```sh
awslocal s3 mb s3://score-set-csv-uploads-dev
```

Floci's `latest-compat` image bundles the standard AWS CLI. Verify whether `awslocal` is available in the chosen Floci image; if not, replace the call with:

```sh
aws --endpoint-url=http://localhost:4566 s3 mb s3://score-set-csv-uploads-dev
```

Consider renaming the script to `floci-init.sh` to match the new service name and update the volume mount path in `docker-compose-dev.yml`.

**`api/settings/.env.template`**

Update the S3 endpoint URL to match the renamed compose service:

```
S3_ENDPOINT_URL=http://floci:4566
```

Also update `settings/.env.dev` (or its generation process) to keep the running dev environment in sync.

**No test changes required.** Tests mock `mavedb.routers.score_sets.s3_client` and `mavedb.worker.jobs.variant_processing.creation.s3_client` directly via `unittest.mock` and do not start or connect to the emulator.

**Known Floci limitation:** Floci has a confirmed bug where `ListObjects` returns results in reverse alphabetical order. MaveDB only calls `PutObject`, `GetObject`, and `DeleteObjects`, so this does not affect any current functionality.

bencap:
## Summary
Replaces LocalStack with moto for local S3 emulation. LocalStack archived its Community edition in March 2026 and moved S3 behind a paid token, so `docker compose up` fails with auth errors for anyone pulling the current image.

## Why moto (not S3Mock or Floci)
- **Adobe S3Mock** is Java-SDK-first: every boto3 `PutObject` fails against it (tested 4.6.0 and 5.2.0) because its expect-100-continue handshake drops the connection. It would require path-style + Expect workarounds in our shared `s3_client()` — which also serves production.
- **moto** (Apache-2.0) is built for boto3 and runs the CSV upload/download flow with an unmodified client. It's also multi-service, leaving room for e.g. local Secrets Manager later without another migration.
- **Floci/MiniStack/kumo** are young multi-service LocalStack clones — more than we need, less proven with boto3.

## Changes
- `docker-compose-dev.yml`: `localstack` → `moto` + one-shot `moto-init` (`curlimages/curl`) that PUTs the upload bucket once moto is up.
- Deleted `bin/localstack-init.sh` (and the `awslocal` dependency).
- `settings/.env.template`: `S3_ENDPOINT_URL=http://moto:5000`.
- Host port published as `5001:5000`.
- No production code changed; unit tests already mock the S3 client directly.

### DIFF
--- a/bin/localstack-init.sh
+++ b/bin/localstack-init.sh
@@ -1,4 +1,0 @@
-#!/bin/sh
-echo "Initializing local S3 bucket..."
-awslocal s3 mb s3://score-set-csv-uploads-dev
-echo "S3 bucket 'score-set-csv-uploads-dev' created."
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -103,17 +103,19 @@ services:
     volumes:
       - mavedb-redis-dev:/data
 
-  localstack:
-    image: localstack/localstack:latest
+  # AWS mock.
+  moto:
+    image: motoserver/moto:latest
     ports:
-      - "4566:4566"
-    env_file:
-      - settings/.env.dev
-    environment:
-      - SERVICES=s3:4566 # We only need S3 for MaveDB
-    volumes:
-      - mavedb-localstack-dev:/var/lib/localstack
-      - "./bin/localstack-init.sh:/etc/localstack/init/ready.d/localstack-init.sh"
+      - "5001:5000"
+
+  moto-init:
+    # One-shot: create the upload bucket once moto is up (moto has no auto-create).
+    # A raw PUT is all it takes; --retry-all-errors waits out moto's startup.
+    image: curlimages/curl:latest
+    depends_on:
+      - moto
+    command: -sf --retry 10 --retry-all-errors --retry-delay 1 -X PUT http://moto:5000/score-set-csv-uploads-dev
 
   seqrepo:
     image: biocommons/seqrepo:2024-12-20
@@ -124,4 +126,3 @@ volumes:
   mavedb-data-dev:
   mavedb-redis-dev:
   mavedb-seqrepo-dev:
-  mavedb-localstack-dev:
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -10,6 +10,7 @@ services:
     depends_on:
       - db
       - redis
+      - moto-init
     env_file:
       - settings/.env.dev
     environment:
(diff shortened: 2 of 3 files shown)


jstone-dev:
In docker-compose-dev.yml, should app and dev signal their dependency on moto-init? Otherwise this worked just fine for me (aside from a failing test that is probably unrelated to the PR).

PR #863, opened 2026-09-02
feat(urns): forward URNs retired by publication to the published record

### LINKED ISSUE #617: Forwarding tmp URN to the new published record
It would be convenient if when publishing a dataset, links to the old tmp URN would automatically resolve to the published record.

This is especially apparent when publishing a score set, where after reloading the page you get a URN not found and have to go hunt for the new link.

davereinhart:
This pull request introduces a robust mechanism for forwarding requests that use "retired" URNs (Uniform Resource Names) to the current, published URNs for datasets, experiments, and score sets. This ensures that links shared before publication continue to work after a record is published and its URN changes. The implementation covers database schema changes, backend logic, API integration, and tests.

**URN Redirects: Database and Model Support**
- Added a new `urn_redirects` table (with migration) and corresponding SQLAlchemy model `UrnRedirect` to track mappings from old (retired) URNs to new (published) URNs. This enables the system to look up and forward requests for outdated URNs. [[1]](diffhunk://#diff-1dee69c6bfc84496a790e1f5b80db14c5711d013abe34d8922780fdb89e20e18R1-R35) [[2]](diffhunk://#diff-aab64b97f8a3087ee53afd0d49ccb725e34921477cd1768aee87794ce50615c5R1-R36)
- Registered the `urn_redirect` model in the models package for application-wide access.

**URN Redirect Logic and Integration**
- Implemented `record_urn_redirect` and related logic in `urn_redirects.py` to record URN changes and determine when a request should be forwarded. Only public, published records are eligible as redirect targets, and only read requests (GET/HEAD) are forwarded.
- Integrated URN redirect recording into the publication flow for experiment sets, experiments, and score sets, so that every time a URN is changed during publication, the redirect is saved. [[1]](diffhunk://#diff-6116c1df8273d3b85cb87b98c7c34378c0b7d110fb3703242f035de6ab1a7822R80) [[2]](diffhunk://#diff-6116c1df8273d3b85cb87b98c7c34378c0b7d110fb3703242f035de6ab1a7822R2567-R2593)

**API and Application Integration**
- Applied the `forward_retired_urns` dependency globally to all FastAPI routes, ensuring that any request using a retired URN is automatically redirected to the current URN if appropriate. [[1]](diffhunk://#diff-7e29c17cea76d8de36feb1232adb56b8df88e0de8d9f20449db13cbaa8e783e4L6-R6) [[2]](diffhunk://#diff-7e29c17cea76d8de36feb1232adb56b8df88e0de8d9f20449db13cbaa8e783e4R40) [[3]](diffhunk://#diff-7e29c17cea76d8de36feb1232adb56b8df88e0de8d9f20449db13cbaa8e783e4L78-R81)

**Testing**
- Added comprehensive tests for the URN redirect functionality, covering correct recording, ignoring of no-op or invalid cases, and the forwarding logic for various scenarios (including sub-resources and variants).URN of an experiment set, an experiment and a score set in place, and refresh_variant_urns rebuilds every variant URN from the score set's. Nothing recorded the old value, so a link already shared to the unpublished record began returning 404 under a name the caller had no way to guess. Reloading a score set page after publishing it was enough to hit this.

Refs: VariantEffect/mavedb-ui#617

### DIFF
--- a/src/mavedb/lib/urn_redirects.py
+++ b/src/mavedb/lib/urn_redirects.py
@@ -1,0 +1,187 @@
+"""Forwarding of URNs that publication has retired.
+
+A dataset is created with a ``tmp:<uuid>`` URN and keeps it until it is published, at which point
+:func:`mavedb.routers.score_sets.publish_score_set` overwrites the URN in place with a permanent one.
+Any link already shared to the unpublished record then stops resolving, which is what
+https://github.com/VariantEffect/mavedb-ui/issues/617 reports: the record is still there, under a name
+the caller has no way to guess.
+
+Publication records what each retired URN became, and every read is checked here before it reaches its
+route. A request naming a retired URN is answered with ``308 Permanent Redirect`` to the same path
+under the record's current URN.
+
+Three limits:
+
+  - Only a public target is forwarded to. A ``Location`` header names the record it points at, and
+    forwarding happens before any route checks a permission, so the header would disclose that URN to
+    an anonymous caller. See _target_is_public.
+  - Forwarding is one hop. Nothing in the application renames a published record, so a chain cannot
+    arise; if one ever could, resolution here would need to follow it.
+  - URNs retired before this was added are unrecoverable. Publication overwrote them and no history of
+    them was kept, so links to records published earlier stay broken.
+"""
+
+import logging
+from typing import Optional
+from urllib.parse import quote
+
+from fastapi import Depends, HTTPException
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+from starlette.requests import Request
+
+from mavedb.deps import get_db
+from mavedb.lib.logging.context import logging_context, save_to_logging_context
+from mavedb.lib.validation.urn_re import (
+    MAVEDB_EXPERIMENT_SET_URN_RE,
+    MAVEDB_EXPERIMENT_URN_RE,
+    MAVEDB_SCORE_SET_URN_RE,
+    MAVEDB_TMP_URN_RE,
+)
+from mavedb.models.experiment import Experiment
+from mavedb.models.experiment_set import ExperimentSet
+from mavedb.models.score_set import ScoreSet
+from mavedb.models.urn_redirect import UrnRedirect
+
+logger = logging.getLogger(__name__)
+
+# Methods whose requests are forwarded. See forward_retired_urns for why a write is not.
+SAFE_METHODS = frozenset({"GET", "HEAD"})
+
+# The record kinds publication renames, recognized by the shape of the URN it gave them. Matched with
+# fullmatch, under which the three patterns are mutually exclusive.
+FORWARDING_TARGET_MODELS = (
+    (MAVEDB_SCORE_SET_URN_RE, ScoreSet),
+    (MAVEDB_EXPERIMENT_URN_RE, Experiment),
+    (MAVEDB_EXPERIMENT_SET_URN_RE, ExperimentSet),
+)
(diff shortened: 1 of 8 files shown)


src/mavedb/routers/score_sets.py
bencap, line 2541: This was a gap preexisting this PR, but reading this function I'm realizing a user could repeatedly publish their score set and it would unconditionally change their URN. We should probably add a block preventing that alongside this change and an associated test.

```
if not item.private:
    logger.info(
        msg="Failed to publish score set; The requested score set has already been published.",
        extra=logging_context(),
    )
    raise HTTPException(status_code=409, detail="This score set has already been published.")
```

src/mavedb/lib/urn_redirects.py
```diff
@@ -0,0 +1,187 @@
...
+    :param db: An active database session.
+    :param urn: The URN a redirect points to.
+    :return: True only if a record under this URN exists and is public.
+    """
+    for urn_re, model in FORWARDING_TARGET_MODELS:
+        if urn_re.fullmatch(urn):
+            private = db.execute(select(model.private).where(model.urn == urn)).scalar_one_or_none()
+            return private is False
+
+    return False
```
bencap: This function is re-deriving 'is this record visible to an anonymous caller', which I think `has_permission(None, entity, Action.READ)` already does for all three possible models here. Using that function in place of this helper would guarantee any future visibility changes would also be reflected here.

src/mavedb/lib/urn_redirects.py
```diff
@@ -0,0 +1,187 @@
...
+    ``308`` rather than ``301``: the redirect is permanent, and 308 forbids a client from rewriting the
+    request to a GET on the way, which is what makes the header safe to emit for any method.
+
+    Only reads are forwarded. What the issue asks for is that shared *links* keep working, and a write
+    is a different proposition: the caller addressed a private draft, and the record now under that URN
+    is published, with different rules and a wider audience. ``POST .../publish`` is the sharp case --
+    an owner is permitted to publish a published score set, so forwarding a stale one would rename a
+    live public record. A write to a retired URN keeps getting the 404 it gets today, which tells the
+    client to look the record up again.
+    """
```
bencap, line 145: Not blocking, but I might try to place design rationale in the commit message or PR description rather than the docstring. Kind of related to our discussion at standup, I've found the AI agents rather like doing this historical narration in their comments, and I've been trying to cut it more aggressively recently so that docstrings are more precise to what the caller would actually need to know about the function. Here, it's well scoped and well named so a caller wouldn't need to know about much besides the first two and maybe part of the last paragraph.

bencap (commented):
Looks great! The one blocking item isn't really introduced by your changes, but is something we should fix while we're working on this file.

src/mavedb/lib/urn_redirects.py
davereinhart: Good point. I've noticed this too and the generated comments always seem relevant enough to keep, but agree some of it can move to the commit messages to avoid this type of bloat. I adjusted this one and will probably try to add something to CLAUDE.md to prevent it going forward.

src/mavedb/lib/urn_redirects.py
davereinhart: Added this and replaced the loop using FORWARDING_TARGET_MODELS with an if statement to keep it more readable and ensure the target entity is properly typed.

bencap (approved):
Thanks for the changes!

PR #866, opened 2026-09-11
Fix the outer source not found error in both Method sources and Threshold sources in creating a new score set and calibration.

### LINKED ISSUE #865: Methods source and threshold source can't search the external source normally
In creating a new score set and calibration, server shows an error when pasting https://pubmed.ncbi.nlm.nih.gov/31892348/ to both the method sources and threshold sources. 

The front end shows no result.

EstelleDa:


bencap (commented):
Looking at the issue this is paired with, this does work but I think the fix is aimed at the wrong layer. `_fetch` is a low level function shared by every Rxiv call site and having it swallow these parse failures into an empty list means genuine upstream errors (like Rxiv going down) now reads as 'not found' rather than an error.

The root cause here is that we have no way to differentiate PubMed and bioRxiv IDs; they're indistinguishable, so we are forced to check both. Because of that `find_generic_article` fans out to every db an identifier is valid for. The solution here solves that by making the lower level `_fetch` function not raise, but I think a more precise solution in the `find_generic_article function` would:
- Perform the fan out
- Catch any errors in the fanned out searches and log them
- Drop the candidate search because of the error

Catching and dropping the candidate errored search would allow the successful PubMed search to be returned.

That does mean an explicit search with `db_name="medRxiv"` would continue to crash `_fetch`. But I think that's a separate and narrower gap than what the issue was intending to solve.

